PR #4, opened 2026-03-24
Initial release

that-guy-wade:
Initial release of the ORO subnet codebase.


shardi-b (changes requested):
## Code Review: Initial Release

### Critical Issues

**1. Logic Bug: `_upload_logs` overwrites all problem statuses to SUCCESS** (`subnet/validator/main.py`)
The `_upload_logs` method reports ALL uploaded problems as `ProblemStatus.SUCCESS` regardless of their actual scoring result. If a problem scored as FAILED but logs were uploaded, its status gets incorrectly overwritten. This directly affects evaluation correctness.

**2. Subprocess resource leak on timeout** (`subnet/validator/main.py` ~line 336)
When `subprocess.TimeoutExpired` fires, the child process (Docker container) is NOT killed. Per Python docs, the child process continues running. Need to catch the exception, kill the process, then check for partial output:
```python
except subprocess.TimeoutExpired:
    proc.kill()
    proc.communicate()
```

**3. Hardcoded default Watchtower token** (`docker-compose.yml`, `.env.example`, `subnet/validator/main.py`)
The Watchtower API token defaults to `oro-watchtower-token` everywhere. Anyone who can reach the Watchtower endpoint can trigger container updates. The `.env.example` ships the actual default value rather than an empty placeholder.

**4. Docker socket mounted into validator container** (`docker-compose.yml`)
`/var/run/docker.sock:/var/run/docker.sock` gives the validator full Docker daemon access. Combined with the fact that the validator downloads and executes miner-submitted code (in a sandbox), any sandbox escape escalates to host root.

### High Severity

**5. `os.chdir()` in `_initialize_scorers` is process-global** (`subnet/validator/progress_reporter.py` ~line 107)
`os.chdir()` affects the entire process. If anything runs concurrently (the method runs from `__init__` which could be called from a background thread), CWD flapping causes failures. Consider using absolute paths instead.

**6. `_eval_voucher` crashes on missing keys** (`src/agent/problem_scorer.py` ~line 318)
Accesses `product["price"]`, `product["shop_id"]`, `voucher["budget"]`, `voucher["threshold"]`, etc. without `.get()`. The `voucher` parameter type allows `None` but there's no null guard before accessing it at line 323. This crashes scoring for the entire problem.

**7. Eval directory permissions set to 0o777** (`subnet/validator/main.py` ~line 1562)
World-readable/writable/executable permissions. Use 0o770 with group matching the sandbox user instead.

**8. GET requests bypass model validation** (`docker/proxy/validate_model.js`)
Non-POST requests to `/inference/*` are forwarded to the Chutes API without any validation, allowing sandbox code to probe the API or access unintended endpoints.

### Medium Severity

**9. `_product_cache` unbounded** (`src/agent/problem_scorer.py` ~line 39)
No size limit or eviction. For large evaluation suites, memory grows without bound. Use `functools.lru_cache` or a bounded dict.

**10. `write_progress` has no file locking** (`src/agent/problem_scorer.py` ~line 387)
If multiple workers write concurrently, JSONL lines can interleave and produce corrupt output.

**11. `get_suite_problems` bypasses `_call_api` pattern** (`subnet/validator/backend_client.py` ~line 503)
Manually calls `get_httpx_client().get()` instead of using `_call_api`, skipping unified error handling. `httpx.ReadTimeout` and other exceptions won't be caught.

**12. Silent `shutil.rmtree` failure** (`subnet/validator/main.py` ~line 729)
Cleanup failures are completely silent (`except Exception: pass`). Log a warning so operators know if disk space is being consumed by orphaned eval directories.

**13. `sys.path.insert` pollution** (`subnet/validator/progress_reporter.py` ~line 110)
Never cleaned up. Every `ProgressReporter` instantiation prepends to `sys.path`.

### Recommendation

Items 1-2 are correctness bugs that should be fixed before merge. Items 3-4 are security concerns worth addressing. The remaining items are quality improvements that could be tracked as follow-up issues.

shardi-b (approved):

PR #25, opened 2026-03-30
fix: include response body in validator error messages

that-guy-wade:
## Description

Validator error logs only show the HTTP status code when the SDK's parsed error detail isn't available (e.g., `"HTTP 503"` instead of the actual error message). This makes it hard for operators to diagnose issues.

## Changes Made

- `_handle_response` falls back to raw response body (truncated to 500 chars) when `parsed.detail` is unavailable
- `_call_api` extracts response body from `UnexpectedStatus` exceptions

## Testing

Tested against staging with real API calls and mock responses covering all fallback paths.

```bash
```

38 passed, 5 skipped.

## Checklist

### DIFF
--- a/subnet/validator/backend_client.py
+++ b/subnet/validator/backend_client.py
@@ -217,10 +217,15 @@ class BackendClient:
                 status_code=status,
             )
 
-        # Error cases - extract detail from parsed error
-        detail = f"HTTP {status}"
+        # Error cases - extract detail from parsed error, fall back to body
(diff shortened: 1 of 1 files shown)


shardi-b (approved):

Clean, well-scoped fix. The fallback chain for error detail extraction is logically correct, byte truncation is safely handled with `errors="replace"`, and `hasattr` guards are appropriately defensive.

PR #29, opened 2026-03-31
fix: default inference stats path for Docker sandbox

that-guy-wade:
## Description

Inference stats (`inference_failure_count`, `inference_total`) were always `null` in evaluation results because the `INFERENCE_STATS_FILE` env var was never passed to Docker sandbox containers. The `ProxyClient` only wrote stats when the env var was explicitly set.

Fix: default to `/app/logs/inference_stats.jsonl` (inside the already-mounted logs volume) so stats are always written. The progress reporter already reads from this path.

## Changes Made

- `src/agent/proxy_client.py`: Default `INFERENCE_STATS_FILE` to `/app/logs/inference_stats.jsonl` instead of requiring the env var

## Testing

38 passed, 5 skipped.

```bash
```

## Checklist

### DIFF
--- a/src/agent/proxy_client.py
+++ b/src/agent/proxy_client.py
@@ -98,9 +98,10 @@ class ProxyClient:
         self.retry_delay = retry_delay
         self.api_key = api_key or os.getenv("CHUTES_ACCESS_TOKEN")
         self.inference_stats = InferenceStats()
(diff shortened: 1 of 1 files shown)


shardi-b (approved):

PR #31, opened 2026-03-31
fix: set inference stats as top-level fields on progress update

that-guy-wade:
## Description

Inference stats were being stuffed into `score_components_summary` dict instead of set as the top-level `inference_failure_count` and `inference_total` fields on `ProblemProgressUpdate`. The backend stores these as top-level fields, so the data was never populated.

## Changes Made

- `subnet/validator/progress_reporter.py`: Set `inference_failure_count` and `inference_total` directly on `ProblemProgressUpdate` constructor instead of nesting in `score_components_summary`

## Testing

38 passed, 5 skipped.

## Checklist

- [x] My changes generate no new warnings or errors
- [x] New and existing unit tests pass locally with my changes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #33, opened 2026-04-01
fix: write inference stats incrementally and set correct output path

that-guy-wade:
## Description

Follow-up to #31 (merged). Inference stats were still null because:
1. Stats were written at process exit via `atexit`, but Docker kills sandbox containers before `atexit` fires
2. `SANDBOX_OUTPUT_FILE` env var wasn't set, so stats went to `/tmp` (tmpfs, destroyed with container)
3. Stats reader matched by `problem_id` but the key wasn't always set

## Changes Made

- `src/agent/proxy_client.py`: Write stats incrementally after each inference call instead of at exit. Removed `atexit` dependency.
- `src/agent/run_sandbox.py`: Set `SANDBOX_OUTPUT_FILE` env var so executor writes stats to `/app/logs/`
- `src/agent/sandbox_executor.py`: Reader takes last line's cumulative totals instead of matching by problem_id
- `subnet/validator/progress_reporter.py`: Same reader fix

## Testing

Verified locally with Docker sandbox — stats file is written with correct cumulative counts after each inference call.

```bash
```

38 passed, 5 skipped.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,5 @@ resources/documents.jsonl
 .specstory/
 CLAUDE.md
 subnet/logs/
+data/single_problem.json
+data/test_agent.py
--- a/src/agent/proxy_client.py
+++ b/src/agent/proxy_client.py
@@ -10,7 +10,6 @@ This module provides a minimal HTTP client that handles:
 All requests go through the proxy service for network isolation.
 """
 
-import atexit
 import json
 import os
 import logging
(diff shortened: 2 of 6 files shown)


shardi-b (commented):
## Code Review

**Overall**: Clean, well-motivated fix. The incremental write approach correctly solves the Docker container kill + `atexit` race condition.

### Bug: Unreachable dead code in `progress_reporter.py`

In `_read_inference_stats`, there are two consecutive return statements:

```python
        return last_failed, last_total
        return 0, 0  # unreachable
```

The old `return 0, 0` was not removed when adding the new return. No runtime impact since `last_failed, last_total` defaults to `(0, 0)` and is always returned first, but the dead line should be removed.

### Minor notes

- `_read_inference_stats` in `sandbox_executor.py` still accepts `problem_id` as a parameter but no longer uses it in the logic. Consider removing it (or keeping if callers pass it positionally).
- File I/O inside `_lock` in `InferenceStats._flush()` is fine for this use case but worth noting — the append is fast enough that lock contention won't matter in practice.

No blocking issues found.

shardi-b (dismissed):
Approved. No blocking issues found during code review. One minor cleanup: remove the unreachable `return 0, 0` in progress_reporter.py._read_inference_stats.

shardi-b (dismissed):
## Code Review

**PR: fix: write inference stats incrementally and set correct output path**

Good fix overall — the root cause analysis is solid (atexit doesn't fire in Docker, missing env var, matching logic). The incremental write approach is the right solution.

### Bug Found

**`subnet/validator/progress_reporter.py` — unreachable dead code / missed deletion (lines ~430-431)**

After the refactor, there are two consecutive return statements:

```python
        except (FileNotFoundError, json.JSONDecodeError, OSError):
            pass
        return last_failed, last_total
        return 0, 0  # <-- unreachable, should be removed
```

The old `return 0, 0` was not removed when the new `return last_failed, last_total` was added. Not a functional bug (correct return executes first), but it's dead code that will confuse readers. Compare with `sandbox_executor.py` where the same refactor was done correctly.

**Fix:** Delete the `return 0, 0` line.

### Everything Else Looks Good

- `proxy_client.py`: Incremental flush under lock is appropriate given call frequency. Graceful handling of missing stats file.
- `run_sandbox.py`: Setting `SANDBOX_OUTPUT_FILE` env var fixes the root cause.
- `sandbox_executor.py`: Last-line-wins logic is correct for cumulative stats.
- Tests are well-structured and cover the key scenarios (cumulative writes, no file, missing env var).

Requesting changes for the dead code removal — trivial fix.

that-guy-wade:
Fixed the dead `return 0, 0` line in `progress_reporter.py`.

Re: the `problem_id` parameter note — `_read_inference_stats` in `sandbox_executor.py` does use `problem_id` for matching (line 158: `if str(entry.get("problem_id")) == str(problem_id)`). This is needed because multiple problems share the same stats file when running in parallel via `execute_problems_parallel`.

shardi-b (approved):

PR #35, opened 2026-04-02
fix: validator digest reporting after Watchtower recreation

that-guy-wade:
## Description

Validator stopped reporting its own image digest in `service_versions` after Watchtower recreated the container. The hostname-based container ID lookup fails because the hostname retains the old container ID which Docker can't resolve.

Fix: inspect the validator image (`ghcr.io/oro-ai/oro/validator:stable`) directly instead of trying to resolve the container by hostname. The image is always available since we're running inside it.

## Root cause

```
hostname inside container: c73223bdf6bb
actual container ID:       7c33eba52373
docker inspect c73223bdf6bb → "No such object"
docker inspect ghcr.io/oro-ai/oro/validator:stable → works
```

## Testing

53 passed, 5 skipped.

## Checklist

### DIFF
--- a/subnet/validator/version_collector.py
+++ b/subnet/validator/version_collector.py
@@ -19,9 +19,7 @@ SERVICE_CONTAINERS = [
 ]
 
 # Sandbox image to inspect (pulled but not always running)
-SANDBOX_IMAGE = os.environ.get(
-    "SANDBOX_IMAGE", "ghcr.io/oro-ai/oro/sandbox:latest"
-)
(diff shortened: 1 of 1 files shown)


shardi-b (approved):

**Review summary:**

- **version_collector.py**: The fix to inspect the validator image directly instead of relying on hostname-based container ID lookup is correct. Watchtower container recreation invalidates the hostname, but the image reference is always resolvable. Good fallback chain (stable → latest → hostname).

- **progress_reporter.py**: Null byte sanitization before DB storage is a solid defensive fix against CharacterNotInRepertoireError.

Both changes are well-scoped, well-documented, and low risk.

PR #38, opened 2026-04-03
fix: skip embedding computation on ground truth match (ORO-669)

that-guy-wade:
## Description

When the agent recommends the exact correct product (ground truth match), `rule_score_reward` returns 1.0 regardless of title similarity. But the encoding still ran — `model.encode()` was called for every problem even on GT matches. On slow validators this wastes ~20s per GT match.

## Changes Made

- Skip the entire title embedding block when `is_ground_truth` is true
- For non-GT, encode the agent's product title once instead of per reward-title iteration

## Issue Link

- Related to: ORO-669

## Testing

Compiles and lints clean. No behavioral change — GT matches still return 1.0, non-GT comparisons produce identical results.

**Test Command(s):**
```bash
python3.11 -m py_compile src/agent/rewards/orm.py
ruff check src/agent/rewards/orm.py
```

## Checklist

## Additional Notes

Data from Yuma's slow production run: 10/14 scored problems were GT matches that ran embeddings unnecessarily. At ~20s per encode, this optimization saves ~200s on that run.

### DIFF
--- a/src/agent/rewards/orm.py
+++ b/src/agent/rewards/orm.py
@@ -37,28 +37,27 @@ def rule_score_reward(product: dict, reward: dict) -> tuple[float, Counter, Coun
 
     is_ground_truth = ground_truth_reward(product, reward) == 1
 
-    # title — use precomputed embeddings if available, otherwise encode both
-    model = _get_sentence_model()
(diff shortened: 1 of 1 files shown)


shardi-b (approved):

PR #50, opened 2026-04-13
fix: inline rate limit delay constant to fix validator crash

that-guy-wade:
## Description

The staging validator is crash-looping because `reasoning_scorer.py` imports `from proxy_client import DEFAULT_RATE_LIMIT_RETRY_DELAY` at module level. The `proxy_client` module only exists inside the sandbox container, not the validator container.

## Changes Made

- Replaced the import with an inlined constant `RATE_LIMIT_RETRY_DELAY = 5`

## Testing

The constant value is `5` (verified from `src/agent/proxy_client.py`). This is a one-line fix to unblock the staging validator.

## Checklist

### DIFF
--- a/src/agent/reasoning_scorer.py
+++ b/src/agent/reasoning_scorer.py
@@ -15,7 +15,8 @@ from typing import Any
 
 import requests
 
-from proxy_client import DEFAULT_RATE_LIMIT_RETRY_DELAY
+# Base delay for rate-limit retries (seconds). Matches ProxyClient convention.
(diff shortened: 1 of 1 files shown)


shardi-b (approved):

Clean fix: inlines the `DEFAULT_RATE_LIMIT_RETRY_DELAY = 5` constant (verified matches `proxy_client.py:29`) to break the cross-container import dependency causing the validator crash loop. No behavioral change.

PR #51, opened 2026-04-13
fix: cap retry backoff at 10s to prevent evaluation stalls

that-guy-wade:
## Description

Exponential backoff with 5s base was growing to 40s per retry, causing evaluations to stall for minutes when Chutes inference is flaky. Caps backoff at 10s for both ProxyClient and reasoning scorer.

Worst case per problem drops from 225s to 110s.

## Changes Made

- `proxy_client.py`: `min(base_delay * (2 ** i), 10)`
- `reasoning_scorer.py`: same cap applied to both retry paths

## Testing

```bash
54 tests passed
```

## Checklist

### DIFF
--- a/src/agent/proxy_client.py
+++ b/src/agent/proxy_client.py
@@ -178,7 +178,7 @@ class ProxyClient:
             if i < self.max_retries - 1:
                 is_rate_limited = response is not None and response.status_code == 429
                 base_delay = self.rate_limit_retry_delay if is_rate_limited else self.retry_delay
-                delay = base_delay * (2 ** i)
+                delay = min(base_delay * (2 ** i), 10)
                 time.sleep(delay)
 
         logger.error(f"Failed {operation_name} after {self.max_retries} retries")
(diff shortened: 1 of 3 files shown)


shardi-b (dismissed):

shardi-b (approved):

**Review Summary:**

1. **proxy_client.py** - Backoff cap via `min(..., 10)` is correct. With `base_delay=5`, delays go from 5→10→10→10 instead of 5→10→20→40. Clean.

2. **reasoning_scorer.py** - Same cap applied to both retry paths. Removing the old `min(attempt, 3)` exponent cap is fine since `min(..., 10)` on the result is a tighter bound anyway (old max was 40s, new max is 10s).

3. **subnet/validator/main.py** - Dropping `reasoning_details` from `score_components` is a sensible payload size reduction. The detailed data already gets uploaded to S3 via `_upload_logs`, so no data loss. Good defensive fix against WAF body size limits.

All three changes are straightforward, well-scoped, and consistent with each other.

PR #65, opened 2026-04-19
feat: validate Chutes token before starting evaluation (ORO-861)

that-guy-wade:
## Description

Before starting the sandbox, validate the miner's Chutes access token by hitting the `/models` endpoint. On auth failure (401/402/403), immediately fail the run with a descriptive reason through the normal `_complete_with_failure` flow.

Previously, invalid tokens (e.g., miner's Chutes account has no credits — HTTP 402) caused agents to silently time out on every inference call, wasting a validator slot for ~30 minutes with no useful output.

## Changes Made

- Add `_validate_chutes_token()` method — lightweight GET to validate the token works
- Call it after token extraction, before starting heartbeat/sandbox
- On 401/402/403: fail run immediately with clear reason ("Miner's Chutes account has no credits")
- On 5xx/timeout: proceed normally (transient, token may still work)

## Issue Link

- Closes: ORO-861

## Testing

Manual verification — the validation call is a simple HTTP GET with error handling.

## Additional Notes

This replaces the backend-side approach (Backend PR #236, closed) which required DB manipulation on the failure path. The validator-side approach is cleaner because it uses the existing `_complete_with_failure` flow — no special cancellation logic needed.

### DIFF
--- a/subnet/validator/main.py
+++ b/subnet/validator/main.py
@@ -632,13 +632,19 @@ class Validator:
         agent_path = None
         workspace_dir = Path(self.config.workspace_dir)
 
-        # Step 0: Verify miner Chutes token is present (before starting heartbeat)
+        # Step 0: Verify miner Chutes token is present and valid
         if not chutes_access_token:
             self._complete_with_failure(
                 eval_run_id,
                 TerminalStatus.FAILED,
                 "Miner has no Chutes token — cannot fund inference",
             )
+            return
+
+        # Validate the token can actually make inference calls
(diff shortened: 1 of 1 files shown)


shardi-b (dismissed):

shardi-b (approved):

**Minor notes (non-blocking):**
- `import requests` inside `_validate_chutes_token` is redundant — `requests` is already imported at module level (line 12). Harmless but unnecessary.
- PR description says 403 should fail the run, but code treats it as inconclusive/transient. The code behavior is actually more correct (403 could be rate-limiting), but the description is slightly misleading.

PR #66, opened 2026-04-19
feat: log all proxy API calls in agent trajectories (ORO-772)

that-guy-wade:
## Description

Agents that bypass the `@Tool` decorator and call `ProxyClient` directly produce incomplete trajectories. This makes trajectory data unusable for training (trajectory distillation, ORO-568) and makes it impossible to fully verify agent behavior.

Adds a `RequestLog` sidecar that records every `ProxyClient.get()` and `.post()` call, following the existing `InferenceStats` pattern.

## Changes Made

- `proxy_client.py`: Add `RequestLog` class, instrument `get()` and `post()` with timing and logging
- `sandbox_executor.py`: Wire `REQUEST_LOG_FILE` env var, read sidecar file after execution, attach `proxy_calls` array to trajectory output
- 10 new tests for `RequestLog` and `_read_request_log`

## Issue Link

- Closes: ORO-772

### Automated Testing

**Test Command(s):**
```bash
python3.11 -m pytest subnet/tests/test_proxy_client.py subnet/tests/test_sandbox_executor.py -v
```

26 tests passing (16 existing + 10 new).

### Manual Testing

N/A — requires sandbox environment. Should be validated by running an agent locally and inspecting the trajectory output for `proxy_calls` in `extra_info`.

## Documentation

- [ ] README updated

## Checklist

## Additional Notes

**Backward compatible:** Output format remains a JSON array of dialogue steps. The `proxy_calls` array is stored inside `extra_info` of the first step, which existing consumers ignore.

**Response truncation:** Responses over 2KB are truncated (length recorded, body omitted). Inference request bodies omit the `messages` array to keep logs small.

**Per-problem isolation:** Each problem gets its own `request_log_{problem_id}.jsonl` file to prevent cross-contamination in parallel runs.

### DIFF
--- a/src/agent/proxy_client.py
+++ b/src/agent/proxy_client.py
@@ -19,6 +19,58 @@ DEFAULT_RETRY_DELAY = 2
 DEFAULT_RATE_LIMIT_RETRY_DELAY = 5
 
 
+class RequestLog:
+    """Thread-safe log of all proxy HTTP calls.
+
+    Appends one JSONL entry per request so data survives process kills.
+    """
+
+    def __init__(self, log_file: Optional[str] = None):
+        self._lock = threading.Lock()
+        self._log_file = log_file
+
+    def record(
+        self,
+        method: str,
+        path: str,
+        params: Optional[Dict] = None,
+        json_data: Optional[Dict] = None,
+        status_code: Optional[int] = None,
+        response_body: object = None,
+        duration_ms: float = 0.0,
+    ) -> None:
+        if not self._log_file:
+            return
+        entry = {
+            "method": method,
+            "path": path,
+            "timestamp": int(time.time() * 1000),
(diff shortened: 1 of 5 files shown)


shardi-b (approved):

**Minor observations (all below threshold):**

1. `_read_request_log` catches `JSONDecodeError` at the outer try/except level — a single malformed JSONL line would stop parsing and drop subsequent entries. Moving the catch inside the for-loop would be more resilient. Low practical risk since the writer always uses `json.dumps`. (60/100)

2. In `post()`, non-200 response bodies are not captured in the log (`result` stays `None`). This appears intentional to keep logs clean. (40/100)

Good design choices: per-problem file isolation, response truncation, transient-error tolerance in Chutes token validation, and thread-safe JSONL append pattern.

PR #74, opened 2026-04-20
fix: reduce search-server JVM heap to prevent OOM

that-guy-wade:
## Summary
- Lower JVM heap from `Xmx8g -Xms4g` to `Xmx4g -Xms2g`
- Production search-server stabilizes at ~2.9GB — previous 4GB initial allocation left only 223MB free on the 7.6GB host
- Frees ~1GB headroom to prevent OOM during 10-worker sandbox evaluations
- Still overridable via `JAVA_OPTS` env var

## Test plan
- [ ] Deploy to staging, verify search latency unchanged
- [ ] Monitor memory on prod after promotion — should see ~1GB more free
- [ ] Verify no search failures during evaluation runs

### DIFF
--- a/docker/search-server/entrypoint.sh
+++ b/docker/search-server/entrypoint.sh
@@ -18,7 +18,7 @@ echo "Java version: $(java -version 2>&1 | head -1)" >&2
 
 # Set JVM memory options for pyserini/Lucene (default: 8GB heap, adjust as needed)
 # Use JAVA_OPTS environment variable if set, otherwise use defaults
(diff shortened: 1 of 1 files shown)


shardi-b (approved):

Single-line config change reducing JVM heap from 8g/4g to 4g/2g — well justified by the 2.9GB steady-state observation. Override via JAVA_OPTS is preserved.

Minor nit (non-blocking): the comment on line 19 still says "default: 8GB heap" — consider updating to match the new 4GB max.

PR #75, opened 2026-04-20
fix: lower JVM heap in docker-compose.yml (completes v1.0.38)

that-guy-wade:
## Summary
The entrypoint.sh change in PR #74 had no effect because docker-compose.yml passes `JAVA_OPTS` explicitly as an env var, which overrides the entrypoint default. This PR updates the compose file to match.

## Changes Made
- `JAVA_OPTS` default in docker-compose.yml: `Xmx8g -Xms4g` → `Xmx4g -Xms2g`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,7 +8,7 @@ services:
     environment:
       - HOST=${HOST:-0.0.0.0}
       - PORT=${PORT:-5632}
-      - JAVA_OPTS=${JAVA_OPTS:--Xmx8g -Xms4g}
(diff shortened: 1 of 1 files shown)


shardi-b (approved):

**Review Summary:**
- 1-line change lowering JVM heap defaults from `Xmx8g/Xms4g` to `Xmx4g/Xms2g` in docker-compose.yml
- Correctly addresses the issue where the compose env var was overriding the entrypoint.sh defaults from PR #74
- Bash variable substitution syntax is correct
- Still allows override via `JAVA_OPTS` env var at runtime

PR #78, opened 2026-04-22
fix: handle unescaped control characters in reasoning judge JSON

that-guy-wade:
## Description

The LLM reasoning judge sometimes returns JSON responses with literal newline characters inside string values (e.g., product IDs split across lines). Python's `json.loads` in strict mode rejects these, causing `parse_judge_response` to fall through to `score=0.0` with the raw JSON stored as the explanation.

This means affected problems get a reasoning score of 0 instead of the actual score (e.g., 0.25), directly impacting the agent's reasoning coefficient and final score.

## Root Cause

Example judge response (note `\n\n` inside product IDs):
```json
{"reasoning_quality": 0.25, "explanation": "...product IDs (4\n\n99256732, 4\n\n999474177)..."}
```

`json.loads` fails with "Invalid control character" and both parse attempts fall through to the 0.0 fallback.

## Changes Made

- Use `json.loads(..., strict=False)` on both parse paths to tolerate unescaped control characters
- Add `re.DOTALL` flag to the regex so it matches JSON spanning multiple lines

## Testing

```python
# Before: json.JSONDecodeError → score=0.0
# After: correctly parses score=0.25
json.loads('{"reasoning_quality": 0.25, "explanation": "IDs (4\n\n99256732)"}', strict=False)
```

## Checklist

### DIFF
--- a/src/agent/reasoning_scorer.py
+++ b/src/agent/reasoning_scorer.py
@@ -272,9 +272,12 @@ def parse_judge_response(response_text: str) -> dict[str, Any]:
     if not response_text:
         return {"score": 0.0, "explanation": ""}
 
-    # Try parsing the whole text as JSON first (no <think> block)
+    # Try parsing the whole text as JSON first (no <think> block).
+    # strict=False tolerates control characters (newlines, tabs) inside
+    # JSON string values — the LLM judge sometimes embeds product IDs
+    # or data that contain literal newlines.
     try:
-        data = json.loads(response_text.strip())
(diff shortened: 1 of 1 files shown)


shardi-b (dismissed):

**Review notes:**
- `strict=False` correctly handles the LLM judge embedding literal newlines in JSON string values
- `re.DOTALL` is the right fix for the regex to match JSON spanning multiple lines
- Both changes are minimal and well-scoped to the problem described
- Good PR description with clear root cause analysis and test case

shardi-b (approved):

PR #79, opened 2026-04-22
fix: pin async-substrate-interface to fix validator crash loop

that-guy-wade:
## Summary
- Pin `async-substrate-interface==1.6.4` in validator Docker requirements
- Fixes crash loop where fresh builds pull a newer version that requires `cyscale` instead of `scalecodec`, causing `ImportError: cannot import name 'ScaleBytes'`

## Test plan
- [x] Verified locally — validator starts and completes evaluations with the pin
- [ ] CI passes

### DIFF
--- a/docker/validator/requirements-base.txt
+++ b/docker/validator/requirements-base.txt
@@ -1,3 +1,4 @@
+async-substrate-interface==1.6.4
 bittensor==10.1.0
 bittensor_wallet==4.0.1
 requests==2.32.5


shardi-b (approved):

**Review notes:**
- Single-line dependency pin in `docker/validator/requirements-base.txt` — straightforward fix
- Clear root cause documented (newer `async-substrate-interface` versions switched from `scalecodec` to `cyscale`, breaking `ScaleBytes` import)
- Pinning to `1.6.4` is the right short-term fix; consider adding a comment in the requirements file noting why it is pinned, for future maintainers

PR #81, opened 2026-04-22
ci: add smoke test after image publish

that-guy-wade:
## Description

Adds a smoke test step to the image publish workflow that runs after images are built and pushed. Catches dependency conflicts and import failures before the image can be promoted to stable.

This would have caught the v1.0.42 crash-loop (scalecodec/cyscale conflict) before it reached production validators.

## Changes Made

- New `smoke-test` job that runs after the `merge` step
- Pulls the freshly published validator and sandbox images
- Verifies each can import their entrypoint module without crashing
- If the smoke test fails, the workflow fails (visible before promoting)

## Checklist


shardi-b (approved):

PR #82, opened 2026-04-22
feat: use uv lockfile for deterministic validator builds

that-guy-wade:
## Summary
- Replace `pip` + `requirements-base.txt` / `requirements.txt` with `uv` + `pyproject.toml` + `uv.lock`
- Every transitive dependency is pinned in the lockfile — fresh builds always get identical packages
- Prevents the `async-substrate-interface` breakage (and any future transitive dep surprises)

## Changes Made
- `docker/validator/Dockerfile` — install `uv`, use `uv sync --frozen` instead of `pip install -r`
- `docker/validator/pyproject.toml` — consolidated all deps with explicit PyTorch CPU index
- `docker/validator/uv.lock` — full lockfile (93 packages pinned)
- `requirements-base.txt` and `requirements.txt` can be removed after merge

## Test plan
- [x] Built locally with `--no-cache` — image starts and claims work successfully
- [ ] CI passes


shardi-b (approved):

Minor suggestions (non-blocking):
- Consider pinning the uv image tag (`ghcr.io/astral-sh/uv:0.7.x` instead of `:latest`) for full build reproducibility
- `torch` is the only unpinned dep in pyproject.toml — adding a version constraint would make intent clearer (lockfile handles it regardless)
- `uv run python` in ENTRYPOINT adds slight overhead; could reference the venv python directly since `uv sync` already created it

PR #84, opened 2026-04-23
fix: give judge ground-truth search results and narrow Rule 5 scope

that-guy-wade:
## Description

Fixes two sources of false-positive "fabricated trajectory" flags from the reasoning judge:

1. The judge could see that the agent called `view_product_information` with a list of product IDs, but had no way to verify those IDs came from an actual `find_product` search. When the search response was truncated (>2KB, which is the common case), the judge flagged legitimate selections as hallucinated.
2. Rule 5 ("Thinking-to-action consistency") was being applied too broadly. In practice, the judge was using it to evaluate whether the final product matched query constraints (price, brand, attributes) — which is already scored separately via `success_rate` / `field_matching`. This produced duplicate penalties and outright errors (e.g. judges miscomputing whether a price was in range).

## Changes Made

- `proxy_client.py`: When a `/search/find_product` response is large enough to be truncated, preserve the returned `product_id` values in `entry["result_product_ids"]`. The full response body is still dropped to keep logs bounded.
- `reasoning_scorer.py`: `_format_proxy_call()` now emits a `returned product_ids:` line under any find_product call that has preserved IDs. Capped at 20 shown with a `+N more` suffix.
- `reasoning_scorer.py`: Rewrote Rule 5 of the judge system prompt to:
  - only ask whether the agent actually made the calls its thinking describes, and
  - explicitly exclude constraint matching (price/brand/attributes/availability) from the judge's scope.
- Added 4 tests covering both changes.

## Issue Link

- Related to: N/A
- Closes: N/A

### Manual Testing

Reproduced the symptom against a real run where the judge flagged two Product problems with "recommended product does not appear in search results." Inspected the raw trajectory and confirmed the product IDs *did* appear in the `view_product_information` call params (which come directly from a preceding `find_product` response). With this change, the `returned product_ids` line on each search call now makes that grounding visible to the judge.

**Test Results:**

```
subnet/tests/test_proxy_client.py .................. 18 passed
subnet/tests/test_reasoning_scorer.py ............ 28 passed
```

### Automated Testing

**Test Command(s):**
```bash
python3 -m pytest subnet/tests/test_proxy_client.py subnet/tests/test_reasoning_scorer.py -v
```

## Documentation

- [ ] README updated

**Documentation Changes:**

N/A — internal validator component.

## Checklist

## Additional Notes

The 0.9-1.0 scoring tier still mentions "applies constraints (budget, voucher rules, product requirements)." That phrasing reinforces constraint-checking behavior and may be worth tightening in a follow-up — I kept this PR scoped to the two specific issues above.

Old trajectories captured before this change do not have `result_product_ids`; the judge will continue to score those without the new context. Only new runs benefit.

### DIFF
--- a/src/agent/proxy_client.py
+++ b/src/agent/proxy_client.py
@@ -61,6 +61,14 @@ class RequestLog:
             if len(body_str) > 2000:
                 entry["response_truncated"] = True
                 entry["response_length"] = len(body_str)
+                if "/search/find_product" in path and isinstance(response_body, list):
+                    # Preserve product IDs so the reasoning judge can verify the
+                    # agent's selections came from actual search results.
+                    entry["result_product_ids"] = [
+                        str(item["product_id"])
+                        for item in response_body
+                        if isinstance(item, dict) and "product_id" in item
+                    ]
             else:
                 entry["response"] = response_body
         with self._lock:
--- a/src/agent/reasoning_scorer.py
+++ b/src/agent/reasoning_scorer.py
@@ -107,9 +107,9 @@ Score 0.6-0.8 — GENUINE reasoning:
 
 Score 0.9-1.0 — STRONG reasoning:
 - All of the above, plus:
-- Agent compares options, applies constraints (budget, voucher rules, product requirements)
-- Explains trade-offs in product selection
+- Agent compares options and explains its selection rationale, citing specific data from proxy results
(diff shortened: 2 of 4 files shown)


shardi-b (dismissed):

shardi-b (approved):

PR #90, opened 2026-04-23
fix: aggregate proxy_calls across steps and restore utilization-based judge routing

that-guy-wade:
## Description

Two related fixes to the reasoning judge.

**1. `format_trajectory_for_judge` was reading `proxy_calls` only from `dialogue[0].extra_info`.** Since trajectory construction distributes proxy calls across every step by timestamp, the judge was seeing a tiny subset of the agent's actual HTTP activity — often just a single step's worth, which looks like "0 inference calls" even for agents that made 20+. That triggered the judge's no-reasoning scoring tier.

Fix: aggregate `proxy_calls` from every step's `extra_info` before building the summary.

**Empirical impact on a 12-trajectory sample:**

| Judge | Original formatter | Fixed formatter | Δ |
|---|---|---|---|
| Qwen3-32B mean | 0.404 | **0.688** | +0.284 |

Same trajectories, identical prompt.

**2. Restore simple utilization-ascending judge routing and re-add MiniMax-M2.5 to the pool.**

The ~0.29 MiniMax bias that previously motivated dropping it was largely an artifact of (1). With the aggregation fix, MiniMax realigns to within 0.05 of Qwen3-32B on the same sample while still offering real load relief when Qwen3-32B is saturated.

Qwen3-235B-Instruct is NOT re-added — its ~0.18 bias persists after the formatter fix.

**Alignment comparison after the formatter fix (12 trajectories):**

| Model | MAD vs Qwen3-32B | bias |
|---|---|---|
| MiniMax-M2.5 | 0.079 | -0.046 |
| Qwen3-235B-Instruct | 0.183 | -0.183 |

## Changes Made

- Aggregate `proxy_calls` across every dialogue step in `format_trajectory_for_judge`
- Simplify `_select_models_by_utilization` back to a flat utilization-ascending sort over `JUDGE_MODELS`
- Re-add `MiniMaxAI/MiniMax-M2.5-TEE` to `JUDGE_MODELS`
- Remove the primary/fallback/throttle-threshold state
- Updated tests for both changes (33 pass)

## Issue Link

- Related to: judge-alignment investigation session
- Closes:

### Manual Testing

Replayed 12 sample trajectories through each judge before and after the formatter fix using the production prompt. Verified:
- Qwen3-32B mean rises from 0.404 → 0.688 with no prompt changes
- MiniMax-M2.5 MAD vs Qwen3-32B drops from ~0.29 (historical) to 0.079
[...]

PR #91, opened 2026-04-23
fix: retry judge on unparseable 200 responses instead of surfacing spurious 0.0

that-guy-wade:
## Description

`parse_judge_response` had a silent ambiguity: when the LLM judge returned HTTP 200 with an empty, truncated, or otherwise unparseable body, it returned `score=0.0` — the same value a judge would return when genuinely scoring a regex agent with 0 inference calls. `score_reasoning_quality` could not tell the two apart, so malformed responses were accepted as "the judge scored 0" and surfaced directly to the leaderboard with no retry and no model rotation.

**Observed impact:** A 50-sample alignment pilot (Qwen3-32B + MiniMax + two DeepSeeks on identical trajectories) flagged one outlier where MiniMax-M2.5 produced an apparent `0.00` score on a trajectory that Qwen3-32B and DeepSeek-V3-0324 both scored at `0.90`. The trajectory had 3 real inference calls and 675 tokens — not a regex agent. Re-judging with MiniMax on the same trajectory returned `0.85`, confirming the earlier `0.00` was an unparseable 200, not a real verdict.

## Changes Made

- `parse_judge_response` now returns a `parsed: bool` alongside `score`/`explanation`. Valid-JSON code paths set `parsed=True`; the empty-text and final-fallthrough paths set `parsed=False`.
- `score_reasoning_quality` checks `parsed` on every 200 response. `parsed=False` is handled like a 502: increment `inference_failed`, rotate to the next least-loaded judge via `_select_models_by_utilization`, exponential backoff, retry up to `max_retries`.
- A legitimate `{"reasoning_quality": 0, ...}` still comes back `parsed=True` and is returned directly — regex-agent detections are unaffected.

## Issue Link

- Related to: 50-sample judge alignment pilot following #90
- Closes:

### Manual Testing

Re-judged the outlier trajectory (av=069e12de-6c6, pid=069dd149-1b2) that produced MiniMax `0.00` in the pilot. After this fix the unparseable first call rotates to the next model (Qwen3-32B or a DeepSeek) and returns a valid score instead of surfacing the spurious 0.

**Test Results:**
36 tests pass (added 5 assertions on the `parsed` flag and 2 new behavior tests — one for rotation on unparseable 200, one confirming legit `0` is preserved without rotation).

### Automated Testing

- `TestParseJudgeResponse.test_legitimate_zero_marked_parsed` — valid `reasoning_quality: 0` is `parsed=True`
- `TestScoreReasoningQuality.test_rotates_on_unparseable_200` — first judge returns empty content, caller rotates and returns score from the next judge
- `TestScoreReasoningQuality.test_keeps_legitimate_zero_without_retry` — valid `reasoning_quality: 0` does NOT trigger a rotation

**Test Command(s):**
```bash
PYTHONPATH=subnet python -m pytest subnet/tests/test_reasoning_scorer.py
```

## Documentation

- [ ] README updated

**Documentation Changes:** Updated `parse_judge_response` docstring to document the `parsed` field and call out that `score=0.0` alone is no longer safe to interpret as a verdict without checking `parsed`.

## Checklist

## Additional Notes

If all rotation attempts in `max_retries` still come back `parsed=False`, `score_reasoning_quality` falls through to the existing "all retries exhausted → 0.0" exit path. That floor is unchanged — the worst case now equals what we used to treat every unparseable response as.

### DIFF
--- a/src/agent/reasoning_scorer.py
+++ b/src/agent/reasoning_scorer.py
@@ -283,10 +283,14 @@ def parse_judge_response(response_text: str) -> dict[str, Any]:
     The full response (think + JSON) is stored as the explanation.
 
     Returns:
-        Dict with 'score' (float 0-1) and 'explanation' (str).
+        Dict with 'score' (float 0-1), 'explanation' (str), and
+        'parsed' (bool). When parsed is False the 0.0 score is a fallback
+        for an unparseable response — callers should treat this as a
+        transient judge failure and retry with another model, not as a
+        legitimate "no reasoning" verdict.
     """
     if not response_text:
-        return {"score": 0.0, "explanation": ""}
+        return {"score": 0.0, "explanation": "", "parsed": False}
 
     # Try parsing the whole text as JSON first (no <think> block).
     # strict=False tolerates control characters (newlines, tabs) inside
@@ -297,7 +301,7 @@ def parse_judge_response(response_text: str) -> dict[str, Any]:
         if isinstance(data, dict) and "reasoning_quality" in data:
             score = max(0.0, min(1.0, float(data["reasoning_quality"])))
             explanation = data.get("explanation", "")
-            return {"score": score, "explanation": explanation}
+            return {"score": score, "explanation": explanation, "parsed": True}
     except (json.JSONDecodeError, ValueError, TypeError):
         pass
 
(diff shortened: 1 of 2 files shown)


shardi-b (dismissed):

shardi-b (approved):

PR #93, opened 2026-04-24
fix: register agent module in sys.modules before exec_module

that-guy-wade:
## Summary

Fixes a sandbox loader bug that breaks any agent using `from __future__ import annotations` together with `@dataclass`.

## The bug

`src/agent/sandbox_executor.py::load_agent_from_file` creates the agent module from an `importlib.util` spec but skips the `sys.modules` registration step. When the agent uses the (perfectly valid) combination of PEP 563 lazy annotations and dataclass:

```python
from __future__ import annotations
from dataclasses import dataclass

@dataclass(frozen=True)
class MyConfig:
    threshold: float = 9.0
```

Python 3.10's `dataclasses._is_type()` tries to resolve string annotations via `sys.modules[cls.__module__].__dict__`. Since we never registered the module, that returns `None`, and the class definition crashes with:

```
AttributeError: 'NoneType' object has no attribute '__dict__'
```

Result: the agent fails to import, `agent_main` never exists, and every problem fails in ~0.18s.

## The fix

One-line fix plus cleanup on failure:

```python
module = importlib.util.module_from_spec(spec)
sys.modules[module_name] = module
try:
    spec.loader.exec_module(module)
except Exception:
    sys.modules.pop(module_name, None)
    raise
```

This matches the documented pattern at https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly.

## Regression test

Added `subnet/tests/fixtures/frozen_dataclass_agent.py` (a minimal agent that uses the broken pattern) and `test_frozen_dataclass_with_pep563_annotations` to lock the behavior in. Verified:
- **Without the fix**: test fails in 0.09s with the exact `NoneType` error
- **With the fix**: test passes; all 11 tests in the file green
- `ruff check` clean

## Risk

Very low. Agents are loaded in per-problem child processes via `multiprocessing.Process`, so `sys.modules` registrations are isolated to that child's lifetime. No sandbox-escape implications (the agent already has full `sys.modules` access). Agents that don't use PEP 563 + `@dataclass` see zero behavior change — they never hit the `_is_type` code path that needs the lookup.

### DIFF
--- a/src/agent/sandbox_executor.py
+++ b/src/agent/sandbox_executor.py
@@ -5,6 +5,7 @@ import json
 import logging
 import multiprocessing
 import os
+import sys
 import time
 from concurrent.futures import (
     ThreadPoolExecutor,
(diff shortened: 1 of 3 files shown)


shardi-b (approved):

PR #94, opened 2026-04-24
fix: skip zero-active-instance models when routing judge calls

that-guy-wade:
## Summary

Chutes descales idle judge models to zero instances. When that happens, the utilization API reports `utilization_current=0.0` (no traffic → no load), and our ascending sort in `_select_models_by_utilization` picks the descaled model *first*. Every call routed to it fails, and those failures count toward the judge-infrastructure-failure threshold in the validator.

Observed during Race 11 (2026-04-24):

- MiniMax successful judge calls on the ORO validator dropped from 126/hr at 19:00 UTC to **0/hr at 20:00 UTC** — Chutes silently descaled mid-race.
- In that same hour the ORO validator logged **296 `Judge call failed`**, **129 `Judge call timed out`**, and **12 `Reasoning judge infrastructure`** run classifications. Race 9 and Race 10 showed none.
- MiniMax's utilization entry shows `instance_count: 0`, `active_instance_count: 0`, `action_taken: "scale_up_candidate"` — so it *is* the canary we can read.

## The fix

`_select_models_by_utilization` now filters out any entry with `active_instance_count == 0` before the ascending sort. It logs which models were skipped in the existing `Judge model order by utilization:` line so the routing decision stays visible.

Conservative choices:
- Models **missing from the API entirely** are kept (not all models are always returned) — permissive fallback.
- Models whose entry is **present but missing the `active_instance_count` field** are kept — permissive fallback for partial Chutes responses.
- If filtering would leave **zero models**, fall back to the full static `JUDGE_MODELS` list (calls will still fail, but behavior stays predictable and callers see their existing retry/rotation path).

## Testing

- All 4 existing `TestSelectModelsByUtilization` tests still pass (sort order, missing-model default, API error, non-200 fallback).
- 3 new tests added:
  - `test_zero_active_instances_excluded` — directly reproduces Race 11. Without the fix it fails with the descaled model still first in the returned order; with the fix it passes.
  - `test_all_zero_active_instances_falls_back_to_static` — every model descaled, we return the static list rather than an empty routing order.
  - `test_missing_active_instance_count_field_treated_as_available` — older/partial Chutes responses don't silently drop models.
- `ruff check` clean.

## Risk

Very low. The only case in which behavior changes from the current code is when a judge model is actively reporting `active_instance_count: 0` on the Chutes utilization API, which is exactly the case we're fixing. Every other code path (missing entry, missing field, API error, all-healthy) behaves identically to before.

### DIFF
--- a/src/agent/reasoning_scorer.py
+++ b/src/agent/reasoning_scorer.py
@@ -44,23 +44,59 @@ CHUTES_UTILIZATION_TIMEOUT = 5
 def _select_models_by_utilization() -> list[str]:
     """Query Chutes utilization API and return JUDGE_MODELS sorted by load.
 
+    Models reported with zero active instances are excluded. Chutes can
+    descale a model to 0 when idle; the utilization API then reports
+    utilization_current=0.0 (no traffic → no load), which would otherwise
+    make the ascending sort prefer it. Every call routed to a 0-instance
+    model fails, and those failures count toward the judge-infra-failure
+    threshold in the validator.
+
     Returns models sorted by utilization_current (ascending = least loaded
-    first). Falls back to the static JUDGE_MODELS list on any failure.
+    first), excluding any with active_instance_count == 0. Falls back to
+    the static JUDGE_MODELS list on any API failure, or if filtering would
+    leave no models.
     """
     try:
         resp = requests.get(CHUTES_UTILIZATION_URL, timeout=CHUTES_UTILIZATION_TIMEOUT)
         if resp.status_code != 200:
             return list(JUDGE_MODELS)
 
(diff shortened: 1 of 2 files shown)


shardi-b (approved):

**Review summary:**
- Logic in `_select_models_by_utilization()` correctly filters descaled models (`active_instance_count == 0`) while being permissive for missing entries/fields
- Safe fallback to static `JUDGE_MODELS` list when all models are filtered out
- Improved logging shows which models were skipped
- 3 new tests cover the fix, edge cases, and backward compatibility
- No behavioral change for the happy path — only the descaled-model routing changes

PR #101, opened 2026-04-28
fix(proxy): default BACKEND_URL/BACKEND_HOST so PR #100 doesn't break validators

that-guy-wade:
## Description

PR #100 added `${BACKEND_URL}` and `${BACKEND_HOST}` envsubst variables to the proxy nginx template. Existing validator docker-compose configs (oro-public and external operator copies) do not set those vars, so envsubst writes empty values. The result is `proxy_set_header Host ;` — invalid nginx config — and the proxy crashloops with `[emerg] invalid number of arguments in "proxy_set_header" directive`. Every inference request returns empty, and every problem fails (0/30 succeeded in local testing against staging).

Watchtower would propagate the failure to every production validator the moment `v1.0.56` is promoted to `stable`: Watchtower pulls the new image but preserves the existing container env, so the new variables stay unset. This is a release blocker for `v1.0.56`.

This PR defaults both variables in the proxy `CMD` so the image is self-contained. Operators can still override via the existing env-var path; defaults match production, so unmodified compose configs keep working after a Watchtower image swap.

## Changes Made

- `docker/proxy/Dockerfile` `CMD` now sets `BACKEND_URL` and `BACKEND_HOST` defaults via shell parameter expansion (`: ${VAR:=default}`) before the existing `envsubst` call:
  - `BACKEND_URL` defaults to `https://api.oroagents.com`
  - `BACKEND_HOST` defaults to `api.oroagents.com`

## Issue Link

- Related to: #100
- Closes:

### Manual Testing

Built the patched image locally and ran two scenarios:

1. **No env vars set** (the production / external-operator scenario):
   ```
   set $backend_models_url https://api.oroagents.com/v1/public/inference/models;
   proxy_set_header Host api.oroagents.com;
   ```
   `nginx -t` returns `syntax is ok`.

2. **Env vars overridden** (staging-style override):
   ```
   BACKEND_URL=https://api.staging.oroagents.com BACKEND_HOST=api.staging.oroagents.com docker run ...
   ```
   ```
   set $backend_models_url https://api.staging.oroagents.com/v1/public/inference/models;
   proxy_set_header Host api.staging.oroagents.com;
   ```

**Test Results:**

Without the patch, the unmodified pre-PR compose produces:
```
[emerg] invalid number of arguments in "proxy_set_header" directive in /etc/nginx/nginx.conf:79
```
and validator runs return 0/30 (all problems fail with `Empty inference response`).

With the patch and no env override, nginx config parses cleanly and the proxy serves requests.

[...]

PR #104, opened 2026-04-28
ci(publish): smoke-test proxy image renders nginx config

that-guy-wade:
## Description

Adds a smoke-test step in `.github/workflows/publish-images.yml` that boots the freshly-built proxy image (with the same env vars `docker-compose.yml` exports) and asserts the container reaches `running`. The step runs before the `merge` job tags `:latest` / `:vN`, so a misconfigured proxy image cannot reach any operator.

## Why this guard

v1.0.56 shipped a regression: a new `${BACKEND_URL}` / `${BACKEND_HOST}` was added to `nginx.conf.template`, but neither the proxy `Dockerfile` CMD nor `docker-compose.yml` provided a value. `envsubst` wrote empty strings, nginx crashed with:

```
[emerg] 1#1: invalid number of arguments in "proxy_set_header" directive in /etc/nginx/nginx.conf:79
```

The proxy crashlooped on staging for hours. Every miner inference call hit a dead proxy and timed out at 300s — looking exactly like the ORO-904 hang signature. v1.0.57 (#101) added shell defaults in the proxy CMD as a hotfix.

This smoke step would have failed v1.0.56's publish workflow before the broken image ever got tagged as `latest`.

## Changes Made

- `.github/workflows/publish-images.yml`: download `digest-proxy-linux-amd64` artifact and run a smoke step that boots the image with the compose-defaulted env vars and checks `docker inspect` state is `running` after 5s.

## Issue Link

- Related to: ORO-948
- Related to: ORO-904 (the hang signature this regression masqueraded as)

### Manual Testing
Reproduced the smoke logic locally against `ghcr.io/oro-ai/oro/proxy:latest` (v1.0.58):

```bash
docker run -d --name proxy-smoke-verify \
  -e SEARCH_SERVER_URL=search-server \
  -e SEARCH_SERVER_PORT=5632 \
  -e CHUTES_HOST=llm.chutes.ai \
  ghcr.io/oro-ai/oro/proxy:latest
sleep 5
docker inspect -f '{{.State.Status}}' proxy-smoke-verify
# → running
```

Also verified that without the env vars (mimicking the v1.0.56 break), the same image exits within 5s with the `[emerg]` error — so the guard correctly detects the failure mode.

**Test Results:**
- v1.0.58 (current latest, has CMD defaults): smoke passes
- Bare image with no env: container exits → smoke would fail the publish

### Automated Testing
The new step itself runs in CI on the next tag push.

**Test Command(s):**
```bash
docker run -d --name proxy-smoke <image> -e SEARCH_SERVER_URL=... -e SEARCH_SERVER_PORT=... -e CHUTES_HOST=...
sleep 5
docker inspect -f '{{.State.Status}}' proxy-smoke   # expect "running"
```

## Documentation

- [ ] README updated

**Documentation Changes:**
None — comment in the workflow step explains the rationale and references the v1.0.56 incident.

## Checklist

## Additional Notes

The env vars passed to the smoke test mirror what `docker-compose.yml` already defaults. If a future PR adds a new `${VAR}` to `nginx.conf.template`, the author has two safe paths:

1. Add the var to `docker-compose.yml` AND mirror it in this smoke step's `-e` flags
2. Add a CMD-level default in `docker/proxy/Dockerfile`

Either keeps the smoke green. If the author does neither, the publish workflow fails before tagging.

### DIFF
--- a/.github/workflows/publish-images.yml
+++ b/.github/workflows/publish-images.yml
@@ -130,6 +130,12 @@ jobs:
           name: digest-sandbox-linux-amd64
           path: /tmp/digests/sandbox
 
+      - name: Download proxy digest (amd64)
+        uses: actions/download-artifact@v4
+        with:
+          name: digest-proxy-linux-amd64
+          path: /tmp/digests/proxy
+
       # Override ENTRYPOINT so argv runs our import check instead of booting
       # a live validator. Validator deps live in a uv-managed venv, so we
       # enter through `uv run` rather than the system python.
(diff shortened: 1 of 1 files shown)


shardi-b (approved):

**Review summary:**
- Smoke test follows the same pattern as existing validator/sandbox smoke tests
- Proper error handling with log dumping and cleanup on failure
- Env vars correctly mirror docker-compose.yml defaults
- Well-commented with clear rationale referencing the v1.0.56 incident
- Purely additive change with no risk to existing workflow steps

PR #103, opened 2026-04-28
feat(validator): report host resource metrics on heartbeat

that-guy-wade:
## Description

Validator now samples host resource utilisation (CPU%, RAM%, disk%, running Docker container count) and attaches it to the heartbeat payload on every \`claim_work\` and per-run heartbeat. The Backend persists the latest sample on the validator row so health dashboards can show real numbers instead of placeholders.

## Changes Made

- New \`subnet/validator/resource_collector.py\` snapshots metrics via \`psutil\` and one \`docker ps -q\` call. Failures collecting any individual metric are swallowed so the heartbeat path can never fail because metrics are unavailable. Disk samples \`\$SANDBOX_DIR\` when set, falls back to \`/\`.
- \`BackendClient.claim_work\` and \`BackendClient.heartbeat\` accept an optional \`resource_metrics\` dict. A shared private \`_build_heartbeat_body\` helper constructs the SDK payload for both endpoints so they stay in sync.
- \`HeartbeatManager\` accepts a \`resource_metrics_provider\` callable so each beat samples fresh metrics. A provider exception degrades to \"no metrics this beat\" rather than failing the heartbeat.
- \`main.py\` wires the collector into the claim loop and the per-run heartbeat manager.
- Bumps \`oro-sdk\` pin to \`1.0.50\` (which exposes the new optional fields on \`HeartbeatRequest\`) and adds \`psutil\` to \`requirements.txt\`.

## Issue Link

- Related to: N/A
- Closes: N/A

### Manual Testing

Ran the new \`resource_collector\` test suite locally — happy path, partial-failure path, empty-docker case, and the \`SANDBOX_DIR\` env override. Also verified that \`HeartbeatManager\` keeps heartbeating when the metrics provider raises.

**Test Results:**

\`\`\`
93 passed in 3.30s
\`\`\`

(Two tests in \`test_auto_update.py\` and \`test_progress_reporter_envelope.py\` need a full bittensor install to import \`validator.main\`; they run in CI.)

### Automated Testing

\`\`\`bash
PYTHONPATH=.:.. python -m pytest tests/validator/ -q
\`\`\`

**Test Command(s):**

See above.

## Documentation

- [ ] README updated
- [ ] Code comments added/updated
- [ ] API documentation updated
- [ ] Configuration documentation updated
[...]

PR #105, opened 2026-04-28
hotfix(validator): regenerate uv.lock with psutil + harden smoke test

that-guy-wade:
## Description

Hotfix for v1.0.59 shipping a broken validator image. PR #103 added `psutil>=5.9.0` to `docker/validator/pyproject.toml` but did not regenerate `uv.lock`. The Dockerfile builds the venv with `uv sync --frozen`, which honors the lockfile and skipped psutil. The container's baked `.venv` had no psutil, so the validator crashed at startup with `ModuleNotFoundError: No module named 'psutil'` and crashlooped after Watchtower deployed the image.

The image smoke step was supposed to catch this but invoked `docker run --entrypoint uv ... run python -c "..."`, which re-resolves deps from `pyproject.toml` at exec time — it transparently pulled psutil from PyPI inside the smoke step, hiding the lockfile drift.

## Changes Made

- `docker/validator/uv.lock`: regenerated via `uv lock`. Adds `psutil 7.2.2` (declared in pyproject.toml since PR #103); incidental refresh bumps `oro-sdk` 1.0.45 → 1.0.50.
- `.github/workflows/publish-images.yml`: validator smoke now invokes `.venv/bin/python -c "import subnet.validator.main"` directly — the same interpreter the production ENTRYPOINT uses. Future drift between `pyproject.toml` and `uv.lock` will fail the smoke step before the `merge` job tags the manifest.

## Issue Link

- Related to: ORO-948
- Fixes the v1.0.59 publish that broke staging (and any operator on `IMAGE_TAG=latest`)

### Manual Testing
Reproduced the v1.0.59 failure mode locally against the broken `latest`:

```bash
docker run --rm --entrypoint .venv/bin/python ghcr.io/oro-ai/oro/validator:latest \
  -c "import subnet.validator.main"
# → ModuleNotFoundError: No module named 'psutil'
```

So the new smoke step would have failed v1.0.59's publish before tagging.

After the hotfix builds, the same command should print `Validator module OK`.

**Test Results:**
- New smoke command catches the broken image (proving the fix works)
- `uv lock` regen output confirms psutil 7.2.2 added: `Added psutil v7.2.2`

### Automated Testing
The new smoke step itself runs in CI on the next tag push (v1.0.60 will be tagged after this merges).

**Test Command(s):**
```bash
cd docker/validator && uv lock
docker run --rm --entrypoint .venv/bin/python <image> -c "import subnet.validator.main"
```

## Documentation

- [ ] README updated

**Documentation Changes:**
None — workflow comment explains the rationale and references the v1.0.59 incident.

## Checklist

## Additional Notes

- The same lock-drift class of bug exists across other deps too — this PR only fixes the immediate broken state. Adding a CI step that runs `uv lock --check` (fails if pyproject.toml and uv.lock disagree) would catch the upstream version of this entirely. Worth filing as a follow-up if we keep hitting it.
- Watchtower on staging and the local validator are crashlooping until v1.0.60 publishes. Plan: merge → tag `v1.0.60` → publish → both auto-recover via Watchtower.

### DIFF
--- a/.github/workflows/publish-images.yml
+++ b/.github/workflows/publish-images.yml
@@ -136,14 +136,18 @@ jobs:
           name: digest-proxy-linux-amd64
           path: /tmp/digests/proxy
 
-      # Override ENTRYPOINT so argv runs our import check instead of booting
-      # a live validator. Validator deps live in a uv-managed venv, so we
-      # enter through `uv run` rather than the system python.
+      # Use the baked-in venv at .venv/bin/python — the same interpreter the
+      # ENTRYPOINT runs in production. Going through `uv run` instead would
+      # re-resolve deps from pyproject.toml at exec time and silently install
+      # anything missing from the image, hiding lockfile drift. v1.0.59 shipped
+      # exactly that bug: psutil was added to pyproject.toml but uv.lock wasn't
+      # regenerated, so `uv sync --frozen` skipped psutil at build time and
+      # the validator ImportError'd at runtime.
       - name: Smoke test validator image
         run: |
           digest=$(ls /tmp/digests/validator | head -1)
-          docker run --rm --entrypoint uv ${{ env.IMAGE_PREFIX }}/validator@sha256:$digest \
-            run python -c "import subnet.validator.main; print('Validator module OK')"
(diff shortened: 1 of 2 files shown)


shardi-b (approved):

PR #99, opened 2026-04-27
refactor(validator): ProgressReporter thin orchestrator + OutputWatcher (ORO-907 PR2)

that-guy-wade:
## Description

PR2 of the ORO-907 stack. Folds in ORO-836. Pure structural refactor — no behavior change beyond what PR1 already shipped.

## Changes Made

- New `subnet/validator/output_watcher.py` (~145 LoC): tails `output.jsonl`, parses envelopes, yields typed `ProblemRecord`. Owns file-position tracking, truncation handling, malformed/unknown-status skipping. No scoring/reasoning/backend reporting.
- `ProgressReporter` delegates to `OutputWatcher`: `_file_position`, inline JSON parsing, and `_parse_envelope` are gone. The class is now a wiring orchestrator (watcher → scorer → reasoning judge → backend reporter).
- `_build_terminal_result` now accepts `ErrorInfo | None` (typed dataclass) instead of dict.
- External API of `ProgressReporter` unchanged — no callsite changes in `subnet/validator/main.py`.

## Issue Link

- Closes: ORO-907
- Closes: ORO-836

### Manual Testing

PR1 already shipped envelope IPC. PR2 is structural — same fixtures pass, same behavior.

**Test Results:** 279 passed (274 baseline + 5 new for `OutputWatcher`).

### Automated Testing

**Test Command(s):**
```bash
.venv/bin/pytest tests/ subnet/tests/ -v \
  --ignore=subnet/tests/validator/test_auto_update.py \
  --ignore=subnet/tests/validator/test_weight_setter.py \
  --ignore=tests/integration/test_sandbox_isolation.py
```

## Documentation

- [ ] README updated
- [x] Code comments added/updated
- [ ] API documentation updated
- [ ] Configuration documentation updated
- [x] Other documentation updated (please specify): N/A — internal refactor

**Documentation Changes:**

N/A

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings or errors
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been published and merged

## Additional Notes

Stacks on PR1 #98 (already merged). Diff over main = PR2-only changes.

PR #98, opened 2026-04-27
feat(sandbox+validator): consolidate envelope IPC (ORO-907 PR1)

that-guy-wade:
## Description

PR1 of the ORO-907 stack. Consolidates per-problem sandbox→validator IPC into a single envelope stream in `output.jsonl`. Validator no longer reads `inference_stats.jsonl`. Failures and timeouts are now visible to the validator instead of inferred from absence at a deadline sweep.

Sidecars (`inference_stats.jsonl`, `request_log_*.jsonl`) remain as sandbox-internal child→parent IPC — only the validator-facing reads are removed.

## Changes Made

- Add `SandboxProblemStatus` StrEnum (`SUCCESS / FAILED / TIMED_OUT`) with subset-guard test against Backend `ProblemStatus`
- Extend `ExecutionResult` with `status` field and structured `ErrorInfo` (`type`, `message`); set on every exit path of `execute_single_problem`
- Capture exception type at source in `_run_in_process` (no string-regex heuristics)
- Rewrite `_format_single_result` to emit a flat envelope `{problem_id, status, execution_time, inference_failure_count, inference_total, error, dialogue}` for every outcome
- `ProgressReporter._read_and_dispatch` parses envelope, coerces status to `SandboxProblemStatus` once at the boundary
- Drop validator-side `_read_inference_stats`; envelope carries inference counts
- Failures/timeouts recorded directly via `_build_terminal_result`, no scoring dispatch
- Single `EnvelopeMeta` dataclass replaces parallel dicts; writes/reads under existing lock
- `_mark_remaining_timed_out` continues to handle the genuinely-never-started case (sandbox death before envelope write)

## Issue Link

- Closes part 1 of: ORO-907

### Manual Testing

Local validator + sandbox against staging suite — confirm `output.jsonl` lines are envelopes; confirm Backend `/progress` updates contain entries for failed/timed-out problems. To follow on staging once merged.

**Test Results:** 274 passed (baseline 272 + 2 new for exception-type accuracy).

### Automated Testing

**Test Command(s):**
```bash
.venv/bin/pytest tests/ subnet/tests/ -v \
  --ignore=subnet/tests/validator/test_auto_update.py \
  --ignore=subnet/tests/validator/test_weight_setter.py \
  --ignore=tests/integration/test_sandbox_isolation.py
```

(Ignored files have pre-existing import / docker-environment failures unrelated to this PR.)

## Documentation

- [ ] README updated
- [x] Code comments added/updated
- [ ] API documentation updated
[...]

PR #107, opened 2026-04-29
fix(validator): split output.jsonl by problem envelope before upload (ORO-988)

that-guy-wade:
## Description

**Urgent fix.** v1.0.60 regression: every prod evaluation run uploads exactly **one** trajectory file to S3 instead of one per problem (~30). Frontend trajectory rendering is broken for every problem in every run on v1.0.60. Stable was rolled back to v1.0.54 at 18:41 UTC; this PR is the fix that re-enables a forward roll.

## Root cause

\`_upload_logs\` parses each line of \`output.jsonl\` and groups by \`problem_id\`. Pre-ORO-907 the sandbox wrote a JSON **list of steps** per line, with \`problem_id\` at \`steps[0].extra_info\`. Post-ORO-907 (PR #98) the sandbox writes a **dict envelope** per line:

\`\`\`json
{\"problem_id\": \"...\", \"status\": \"...\", \"dialogue\": [steps], ...}
\`\`\`

The parser still expected the list shape:

\`\`\`python
if isinstance(steps, list) and steps:
    pid = steps[0].get(\"extra_info\", {}).get(\"problem_id\")
\`\`\`

— never matched, \`problem_lines\` stayed empty, fell through to the \"upload entire file under \`problem_ids[0]\`\" fallback, and the resulting S3 object contained the whole multi-envelope JSONL keyed incorrectly under the first problem. Frontend's \`Trajectory = TrajectoryStep[]\` parser also chokes on multi-envelope JSONL. Net effect: trajectory rendering broken for every problem in every run.

This was hiding inside ORO-907 (PRs #98 + #99) for v1.0.56 → v1.0.59 because those tags only saw production traffic via the v1.0.60 promote today.

## Changes Made

- **\`subnet/validator/main.py\`** — extracted the parsing block into a module-level \`_split_output_by_problem\` helper:
  - Recognizes the post-ORO-907 envelope, pulls \`problem_id\` from the envelope itself
  - Uses the envelope's \`dialogue\` array as the upload payload (matching Frontend \`Trajectory = TrajectoryStep[]\`)
  - Empty \`dialogue\` (sandbox crash before agent emitted any steps) still produces a per-problem upload with payload \`[]\` — Frontend renders \"no trajectory\" for that specific problem instead of nothing
  - Pre-907 legacy list format dropped — every code path that reaches the upload step has already gone through \`OutputWatcher\`, which only yields envelope-format records. No real callers produce list-shaped lines.
  - Empty-file fallback (\`problem_ids[0]\` keys whatever's in the file) preserved as a forensic safety net
- **\`tests/test_upload_logs.py\`** — new file covering the parser directly:
  - 3 envelopes → 3 entries (the exact regression scenario)
  - mixed SUCCESS / FAILED / TIMED_OUT terminal statuses all produce uploads
  - unparseable line skipped without losing the good ones around it
  - empty \`dialogue\` yields \`[]\` payload (still uploaded)
  - empty file falls back to \`problem_ids[0]\`
  - no \`problem_ids\` → empty result (sanity)
  - Helper imported via source-level exec to bypass bittensor dep init in the test venv

## No data lost vs the v1.0.54 baseline

The envelope's \`status\` / \`execution_time\` / \`inference_failure_count\` / \`inference_total\` / \`error\` fields are not in the uploaded artifact post-fix. **They were never in the v1.0.54 artifact either** — they flow back via the progress-report path (\`ProgressReporter\` → Backend → eval-run detail endpoint). The artifact has always been just the trajectory steps. This restores Frontend behavior to what worked in v1.0.54.

## Issue Link

- Closes: [ORO-988](https://linear.app/oroai/issue/ORO-988)
- Parent: [ORO-907](https://linear.app/oroai/issue/ORO-907) (envelope IPC consolidation; PRs #98 + #99 are where this regression was introduced)
- Related: PR #98 (sandbox+validator envelope IPC consolidation), PR #99 (ProgressReporter thin orchestrator + OutputWatcher)

### Manual Testing

Locally:

\`\`\`bash
$ uv venv && uv pip install pytest pytest-asyncio
$ .venv/bin/python -m pytest tests/ --ignore=tests/integration -v
... 64 passed in 7.27s
\`\`\`

The new test file's first case (\`test_three_envelopes_yield_three_entries\`) is the exact regression — fails on \`origin/main\` (returns 1 entry under \`problem_ids[0]\`), passes after this fix (returns 3 entries each keyed by their own envelope's \`problem_id\`, with \`dialogue\` arrays as payloads).

End-to-end repro / verification path (recommended pre-promote):

1. Build the validator + sandbox images from this branch locally.
2. Run a 3-problem suite via \`docker compose run test --agent-file <agent> --problem-file <suite>\`.
3. Confirm \`output.jsonl\` has 3 envelope lines.
4. Run the validator against a stub Backend (or staging) and observe **3** \`Generated upload URL\` log lines, **3** S3 keys (one per problem_id), each containing a JSON-array trajectory.

### Automated Testing

64 unit tests pass.

**Test Command(s):**

\`\`\`bash
.venv/bin/python -m pytest tests/ --ignore=tests/integration -v
\`\`\`

## Documentation

- [ ] README updated — N/A
- [x] Code comments added/updated — \`_split_output_by_problem\` has a full docstring explaining the envelope shape, the dialogue-vs-envelope distinction, and the fallback semantics
- [ ] API documentation updated — N/A
- [ ] Configuration documentation updated — N/A
- [ ] Other documentation updated — see Additional Notes for changelog

**Documentation Changes:** changelog entry to record the v1.0.60 → v1.0.54 hop and this fix should land separately (AC item).

## Checklist

- [x] Any dependent changes have been published and merged — depends on PRs #98 + #99 which are already on main

## Additional Notes

**Followups still open from the AC (not in this PR):**

1. **CI smoke test** that runs a tiny eval cycle against a stub Backend and counts upload URL requests, mirroring the proxy nginx + cpu-shares smoke tests added in PRs #102/#104. Worth doing — the unit test caught the regression but a full-stack smoke test would have caught the v1.0.60 release blocker before promote.
2. **Changelog entry** for the v1.0.60 → v1.0.54 rollback + this re-roll. Not in this PR; will land alongside the version bump.
3. **Verify scoring data path is unaffected.** The bug only affected the trajectory-upload artifact. Scoring uses \`OutputWatcher → ProgressReporter → ProblemScorer\` which already consumed the envelope format correctly; spot-checked that \`ProblemScorer\` reads \`record.dialogue\` not the artifact in S3. Recommend a quick code-walk before re-promote anyway.

After merge: re-tag, run the publish-images workflow, verify the smoke tests pass on the new tag, then promote stable from \`latest\` per \`feedback_promote_from_latest\`.


shardi-b (approved):

PR #114, opened 2026-05-01
fix(judge): blacklist judge model after unparseable response, no backoff

that-guy-wade:
## Description

Stops the validator's reasoning judge from burning its full retry budget on a Chutes TEE instance that is serving HTTP 200 with empty `content` on every call.

Today's prod logs (single validator, 60 min window):

| Model | Empty-content responses |
|---|---|
| `moonshotai/Kimi-K2.5-TEE` | ~141 |
| `zai-org/GLM-5-TEE` | ~56 |
| `zai-org/GLM-5.1-TEE` | ~14 |
| `google/gemma-4-31B-turbo-TEE` | 0 |

Failure rate hit 56–63% per eval, which trips `Reasoning judge infrastructure failure` and flips the eval to retry on another validator. The broken instances still report a healthy `active_instance_count` in the Chutes utilization API, so `_select_models_by_utilization` keeps routing to them.

## Changes Made

- Track a per-eval `bad_models` set in `score_reasoning_quality`; once a model returns an unparseable 200 it is skipped for the rest of this eval and the loop aborts early when every model in the rotation is blacklisted, instead of spinning through the remaining retry budget on already-broken models.
- Drop the rate-limit exponential backoff on the unparseable-200 path. Empty content is a model-health failure, not a rate limit — sleeping before rotating just compounds with the per-call timeout and pushes evals past the 900s scoring window. 429/5xx responses still go through `_rotate_and_backoff` unchanged.

## Issue Link

- Related to: N/A (operational fix surfaced from prod logs 2026-04-30 / 2026-05-01 UTC)
- Closes: N/A

### Manual Testing

Re-ran `pytest subnet/tests/test_reasoning_scorer.py -v` — all 44 tests pass, including the three new tests covering the blacklist + no-backoff behavior. Existing `test_returns_zero_when_all_retries_unparseable` continues to pass under the new abort-early path because the rotation cycles through every model before the abort condition is reached at `max_retries=3`.

**Test Results:**
44 passed in 2.45s.

### Automated Testing

New tests:
- `test_blacklists_model_after_unparseable_response` — first attempt empty, second attempt must use a different model and return its score.
- `test_aborts_when_all_models_blacklisted` — `max_retries=20` but only `len(JUDGE_MODELS)` attempts are made before abort.
- `test_no_backoff_sleep_on_unparseable_response` — `time.sleep` is never called on the unparseable-200 path.

**Test Command(s):**
```bash
pytest subnet/tests/test_reasoning_scorer.py -v
```

## Documentation

- [ ] README updated

**Documentation Changes:**
Inline comments in `score_reasoning_quality` explain why empty-content responses are blacklisted per-eval and why the rate-limit backoff is skipped on this path.

## Checklist

## Additional Notes

The Chutes team has been pinged separately about the empty-content responses on Kimi-K2.5-TEE / GLM-5-TEE / GLM-5.1-TEE. This PR is the validator-side defence so a single broken upstream instance doesn't cascade into eval failures across the network.

### DIFF
--- a/src/agent/reasoning_scorer.py
+++ b/src/agent/reasoning_scorer.py
@@ -614,10 +614,24 @@ def score_reasoning_quality(
     headers = {"Authorization": f"Bearer {api_key}"}
 
     models = _select_models_by_utilization()
+    # Models that returned an unparseable 200 in this eval. Skipped on
+    # subsequent rotation steps so a single broken model (e.g. Chutes
+    # serving empty `content` for one TEE variant while reporting healthy
+    # active_instance_count) doesn't burn the full retry budget.
+    bad_models: set[str] = set()
     model_idx = 0
     inference_failed = 0
     inference_total = 0
     for attempt in range(max_retries):
+        if len(bad_models) >= len(models):
+            logger.error(
+                f"All {len(models)} judge models returned unparseable responses "
+                f"this eval; aborting after {attempt} attempts"
+            )
+            break
+        # Skip blacklisted models without consuming a retry slot.
+        while models[model_idx % len(models)] in bad_models:
+            model_idx += 1
         model = models[model_idx % len(models)]
         inference_total += 1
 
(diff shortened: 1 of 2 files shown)


shardi-b (approved):

**Review summary:**
- Per-eval `bad_models` blacklist correctly prevents retry budget from being burned on broken upstream models
- Guard before the skip-`while`-loop prevents infinite loop when all models are blacklisted
- Removing backoff on unparseable 200s is the right call — these are model-health failures, not rate limits
- Three new tests thoroughly cover the blacklist, abort-early, and no-backoff behaviors
- Existing 429/5xx backoff paths are unchanged
- Clean, minimal diff with no unnecessary changes

PR #119, opened 2026-05-01
fix(proxy): serve stale allowlist on Backend fetch failure

that-guy-wade:
## Description

The validator's inference proxy was failing closed (`503 Inference allowlist unavailable`) whenever a post-cache-expiry refresh of `/v1/public/inference/models` returned non-200. Today's prod incident showed this turning a transient Backend 429 (global per-IP rate limit, 100/min) into a 503 storm: every inference request out of the validator was rejected for several minutes, agent evaluations collapsed mid-run, and the validator's reasoning judge — which routes through the same proxy — timed out enough to flip evaluations to STALE / FAILED across multiple validators.

## Changes Made

- `docker/proxy/validate_model.js` keeps the previously-cached allowlist for `STALE_GRACE_MS` (1h) beyond the regular `CACHE_TTL_MS` expiry. If a refresh fails inside the grace window, the proxy serves the stale list and logs the remaining grace time. Outside the grace window we still fail closed, so a genuine prolonged Backend outage is still surfaced.
-
-

## Issue Link

- Related to: N/A (operational fix surfaced from prod logs 2026-05-01)
- Closes: N/A

### Manual Testing
The change is in nginx-njs JavaScript with no test harness in this repo. Verified manually by reading the diff and tracing the cache state machine against today's incident logs.

**Test Results:**
- Cache fresh → callback with cachedList (unchanged)
- Cache expired + refresh 200 → callback with new list, expiry + grace bumped (unchanged)
- Cache expired + refresh 429/5xx + Date.now() < cacheStaleUntil → callback with stale cachedList, error logged (NEW)
- Cache expired + refresh 429/5xx + Date.now() >= cacheStaleUntil → callback null → 503 (unchanged for prolonged outage)
- No prior cache (cold start) + refresh fails → callback null → 503 (unchanged, can't serve what we never had)

### Automated Testing
N/A — no JS tests in this repo.

**Test Command(s):**
```bash
# n/a
```

## Documentation

- [ ] README updated
- [x] Code comments added/updated
- [ ] API documentation updated
- [ ] Configuration documentation updated
- [ ] Other documentation updated (please specify): N/A

**Documentation Changes:**
Updated the file header comment in `validate_model.js` to describe the stale-cache fallback policy and explain why it matters (the prod incident).

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings or errors
[...]

PR #121, opened 2026-05-01
fix(proxy): share inference-allowlist cache across nginx workers

that-guy-wade:
## Description

The per-worker njs cache for the inference allowlist was effectively useless. nginx runs 8 worker processes (one per CPU) and each holds its own njs module state, so all 8 workers cold-start with an empty cache. On startup they each issue `GET /v1/public/inference/models` to the Backend within the same second; combined with the validator's normal heartbeat / weight-setter traffic, those 8 fetches push the validator host past Backend's 100/min global IP rate limit and most workers receive 429. Workers that get 429 never populate their cache, miss on every subsequent inference request, retry the fetch, get 429 again, and answer the client with 503. The earlier `STALE_GRACE_MS` fallback (PR #119) didn't help because it only triggers when there *was* a previous cached value, which these workers never had.

Prod data after v1.0.69 rolled (proxy started 2026-05-01 17:57:15, sample at 14 min uptime):
- 753 Backend models fetch attempts, every one 429
- 753 503s back to inference clients
- 0 "Serving stale allowlist" log lines (no fallback path possible — no prior cache)

This PR moves the cache into a shared-dict zone (`oro_models`, declared in `nginx.conf.template`). Any single successful fetch warms the cache for every worker; subsequent cold-start workers read the shared value instead of issuing their own fetch. Cold-start storms can still race for that first fill, but once one worker wins the rest serve from cache. After a refresh fails, we re-read the shared state before falling back to stale, so a worker stuck on its own failing subrequest can still pick up a fresh value another worker committed mid-flight.

## Changes Made

- Add `js_shared_dict_zone zone=oro_models:64K timeout=24h;` to `docker/proxy/nginx.conf.template`.
- Replace per-worker `var cachedList`, `var cacheExpiresAt`, `var cacheStaleUntil` with shared-dict reads/writes (`ngx.shared.oro_models`).
- Re-read the shared state inside the failure branch so a worker waiting on a failing fetch can still pick up a value committed by a peer worker.

## Issue Link

- Related to: N/A (operational follow-up to PR #119)
- Closes: N/A

### Manual Testing
No JS test harness in this repo. Verified manually:

1. Cache fresh (Date.now() < state.expiresAt) → callback(state.allowlist), no fetch.
2. Cache empty + fetch 200 → write {allowlist, expiresAt, staleUntil}, callback(allowlist).
3. Cache empty + fetch 429/5xx + peer worker just succeeded → re-read finds fresh state → callback(state.allowlist).
4. Cache empty + fetch 429/5xx + only stale state available → callback(state.allowlist), error logged with grace remaining.
5. Cache empty + fetch 429/5xx + nothing in shared dict at all → callback(null) → 503.

After deploy on prod we should see backend `429` count drop from ~50/min to ~0 once any worker successfully fills the shared cache.

### Automated Testing
N/A — no JS tests in this repo.

**Test Command(s):**
```bash
# n/a
```

## Documentation

- [ ] README updated

**Documentation Changes:**
File header in `validate_model.js` rewritten to describe why the cache is shared across workers (the cold-start storm explanation). Inline comment on the `js_shared_dict_zone` directive in `nginx.conf.template`.

## Checklist

## Additional Notes

This builds on PR #119 (stale-cache fallback). After merge, **must burn in on staging first** before promoting to stable; the previous proxy bump skipped staging and Seth correctly flagged that as a violation. Tag and publish via `:latest`, leave staging alone for a few cycles, then promote separately on explicit approval.

### DIFF
--- a/docker/proxy/nginx.conf.template
+++ b/docker/proxy/nginx.conf.template
@@ -6,6 +6,14 @@ http {
     # Load njs validation script
     js_import validate_model from /etc/nginx/validate_model.js;
 
+    # Cross-worker shared cache for the inference allowlist. njs module-level
+    # vars are per-worker, so 8 workers each cold-start with an empty cache
+    # and 8 simultaneous fetches storm the Backend, where some get 429 and
+    # those workers never warm up. A shared zone lets one successful fetch
+    # serve every worker. 64K holds well under 1KB of state today; the
+    # headroom is for future entries.
+    js_shared_dict_zone zone=oro_models:64K timeout=24h;
+
     # Limit request body size (prevents memory exhaustion from oversized payloads)
     client_max_body_size 10m;
 
@@ -76,7 +84,7 @@ http {
 
         # Internal location: fetch the allowed-models list from the ORO Backend
         # so the proxy stays in sync without an image rebuild. njs subrequests
-        # this on cache miss; per-worker cache TTL lives in validate_model.js.
+        # this on cache miss; shared-cache TTL lives in validate_model.js.
         location = /_backend_models {
             internal;
 
--- a/docker/proxy/validate_model.js
+++ b/docker/proxy/validate_model.js
@@ -2,26 +2,49 @@
 // allowlist before forwarding to Chutes.
 //
 // The allowlist is fetched from the ORO Backend (`GET /v1/public/inference/models`)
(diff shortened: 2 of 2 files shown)


shardi-b (approved):

PR #131, opened 2026-05-05
fix(validator): exact t_top share + remove fallback path (ORO-1008)

that-guy-wade:
## Description

Two finishing changes for the top-50% weight distribution after staging-shadow logs validated the algorithm against real race data:

1. **Exact t_top share.** When protected finishers are deregistered between race close and weight set, their tail u16 entries vanish from the submitted vector. Previously top_u16 / burn_u16 were derived from the assumed full tail_sum, so their normalised shares inflated proportionally (race 20 saw top at 25.99% instead of 25.00%). Recompute pinned weights from the **post-filter** tail_sum so the top miner lands at exactly t_top regardless of dereg count. K (size of the protected set) is still derived from the full finisher count, preserving the original protection semantics.

2. **Remove fallback / shadow.** The legacy top-miner fallback, the shadow-mode flag, and the ORO_WEIGHT_SETTER_LIVE env var are gone. Race-based path is the only path. When no completed race is available (fresh subnet, backend outage), skip the tick and let the next one retry — no more silently submitting a stale or non-race vector.

## Changes Made

- weight_distribution.build_metagraph_weight_vector: filter into metagraph first, sum tail_sum_actual, then call compute_pinned_weights to derive top_u16 / burn_u16 from the actual placed tail. Top miner deregistered → submit only burn slot.
- weight_setter._tick: race-only path. No more top-miner fetch, no shadow log branch. Backend error or no-completed-race → skip + warn.
- Drop _build_weights_top_miner_fallback, _log_shadow_race_vector, shadow_mode parameter.
- main.py: drop ORO_WEIGHT_SETTER_LIVE env var read.
- Tests: removed all fallback / shadow / top-miner cases. Added test_drift_correction_when_protected_finishers_deregistered which verifies top share = exactly 25% even with all tail entries dereg'd.

## Issue Link

- Related to: ORO-1008
- Closes: ORO-1008

### Manual Testing
After v1.0.79 deploys to staging:
- Boot log: `Weight setter started (interval: 300s)` (no shadow_mode mention).
- Each tick: `Race-based weight vector: N=N finishers, M non-zero metagraph slots` then `Successfully set weights`.
- On-chain: top u16 share equal to t_top of total submitted u16 (within rounding).
- No more `[SHADOW]` lines, no more `Using top-miner weight path` lines.

**Test Results:**
Pending staging deploy.

### Automated Testing
44 weight tests pass.

**Test Command(s):**
\`\`\`bash
.venv/bin/python -m pytest subnet/tests/validator/test_weight_setter.py subnet/tests/validator/test_weight_distribution.py -q
\`\`\`

## Documentation

- [ ] README updated

**Documentation Changes:**
Inline comments on the post-filter recompute and on `_tick`'s skip-on-no-race branch.

## Checklist

## Additional Notes

After this lands + v1.0.79 promotes, prod will switch from the old 25/75 fallback to race-based weights for the first time. Confidence comes from the staging-shadow comparison against race 20 where the algorithm produced the expected vector across both ORO instances byte-identically.

### DIFF
--- a/subnet/tests/validator/test_weight_setter.py
+++ b/subnet/tests/validator/test_weight_setter.py
@@ -1,18 +1,14 @@
 import time
-from datetime import datetime
 from unittest.mock import MagicMock
 from uuid import UUID, uuid4
 
 import pytest
-
-from oro_sdk.models import TopAgentResponse
 
 from validator.weight_distribution import compute_pinned_weights
 from validator.weight_setter import WeightSetterThread
 
 
 def _empty_history():
-    """A `get_race_history` response with no races — drives the fallback path."""
     history = MagicMock()
     history.races = []
     return history
@@ -57,15 +53,8 @@ def _race_detail(qualifiers: list[dict]):
 
 
 class TestWeightSetterThread:
-    """Integration tests for WeightSetterThread.
-
-    Uses fixtures from conftest.py.
-    """
-
     @pytest.fixture
     def mock_backend_client(self, mock_backend_client_with_top_miner):
-        """Top-miner fixture extended with an empty race history so tests
-        that don't set up a race fall through to the fallback path."""
         mock_backend_client_with_top_miner.get_race_history.return_value = (
             _empty_history()
         )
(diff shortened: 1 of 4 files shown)


shardi-b (approved):

PR #135, opened 2026-05-07
fix(validator): null reasoning_score when judge handed an empty dialogue

that-guy-wade:
## Description

`score_reasoning_quality` in `src/agent/reasoning_scorer.py` returns a sentinel `{score: 0.0, model: "", inference_total: 0}` when called with an empty dialogue (or one that the judge formatter strips to empty). The validator currently records that `0.0` as a real reasoning score and feeds it into the miner's reasoning-coefficient aggregation, which is unfair — the judge never actually rated the trajectory, but the miner's `final_score = success_rate × reasoning_coefficient` gets dragged down as if it had.

Detect the sentinel in `_run_reasoning_judge` and return the existing `empty` dict (`reasoning_score=None`) so the problem is excluded from aggregation. This mirrors the treatment already used when the judge circuit breaker is open or no Chutes token is available.

## Changes Made

- `subnet/validator/progress_reporter.py`: in `_run_reasoning_judge`, after the call to `score_reasoning_quality`, check for the empty-sentinel signature (`score=0.0 AND model="" AND inference_total=0`) and short-circuit to the `empty` return dict with a warning log line. A real judged zero has `inference_total >= 1` and a non-empty model, so it is preserved as-is.
- `subnet/tests/validator/test_progress_reporter_envelope.py`: three new test cases covering (a) empty sentinel → null score, (b) real judged zero preserved, (c) normal judged result preserved.

## Issue Link

- Related to: N/A
- Closes: N/A

### Manual Testing
N/A — pure unit-test change. The behavior under test fires only when `score_reasoning_quality` returns the empty sentinel, which is the exact case the patch handles.

**Test Results:**
13 tests in `test_progress_reporter_envelope.py` pass (the 10 existing + 3 new). Full `test_reasoning_scorer.py` (48 tests) also clean.

### Automated Testing

**Test Command(s):**
```bash
.venv/bin/pytest subnet/tests/validator/test_progress_reporter_envelope.py subnet/tests/test_reasoning_scorer.py -v
```

## Documentation

- [ ] README updated

**Documentation Changes:**
Inline comment block at the new branch in `_run_reasoning_judge` explains the sentinel discriminator and why the patch returns `empty` instead of the judge result. No external docs touched.

## Checklist

## Additional Notes

Discriminator rationale: the judge formatter at `src/agent/reasoning_scorer.py:617-623` returns the empty sentinel only when `dialogue` is falsy or `format_trajectory_for_judge(dialogue)` returns an empty string — both happen *before* any inference call, so `inference_total` is guaranteed to be 0. A real judge run that scored the trajectory zero has `inference_total >= 1` and a model name set, so the discriminator cleanly separates "skipped" from "scored zero".

This patch addresses the fairness half of a recurring upload/dialogue-loss pattern. The second half — making the per-problem trajectory artifact retrievable when this fires — is upload-side and out of scope for this PR.

### DIFF
--- a/subnet/tests/validator/test_progress_reporter_envelope.py
+++ b/subnet/tests/validator/test_progress_reporter_envelope.py
@@ -145,8 +145,13 @@ class TestEnvelopeParsing:
         out = tmp_path / "output.jsonl"
         with open(out, "a") as f:
             f.write("not json\n")
-        _write_envelope(out, problem_id=_P1, status="FAILED", dialogue=None,
-                        error={"type": "RuntimeError", "message": "x"})
+        _write_envelope(
+            out,
+            problem_id=_P1,
+            status="FAILED",
+            dialogue=None,
+            error={"type": "RuntimeError", "message": "x"},
+        )
         reporter._read_and_dispatch()
         assert reporter._results[_P1].status == ProblemStatus.FAILED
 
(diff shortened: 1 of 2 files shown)


shardi-b (approved):

that-guy-wade:
Closing — deferring this fix. Recurrence rate is low (~2 hits in the last 14 days, sample of top 30 agents) and the empty-sentinel discriminator added complexity to a hot path that I'd rather not carry without a clearer scoring-impact signal first.

Will reopen if the rate climbs.

PR #141, opened 2026-05-07
feat(judge): consume Backend ranked endpoint instead of polling Chutes directly

that-guy-wade:
## Description

The judge's model-rotation order used to come from per-validator polls of Chutes' utilization API, with a 30s local cache and per-provider sort logic baked into `reasoning_scorer.py`. OpenRouter had no equivalent signal so it fell back to a static list — asymmetric across providers.

The Backend now polls every upstream provider once per instance and serves a ranked snapshot at `GET /v1/public/inference/models?ranked=true` with provider-specific stats already sorted server-side. This PR moves the judge to consume that endpoint.

Net effect: every upstream stats API gets hit once per Backend instance instead of once per validator, the per-provider sort logic stays server-side where it belongs, and adding a third provider tomorrow means writing one Backend poller class — no validator changes.

## Changes Made

- `src/agent/reasoning_scorer.py`:
  - Replace `_select_models_for_provider`, `CHUTES_UTILIZATION_URL`, and the 30s `_model_order_cache` with `_fetch_ranked_models(provider, backend_url)` that calls the Backend's ranked endpoint
  - Falls back to `JUDGE_MODELS_BY_PROVIDER` static list on any failure (no `backend_url`, network error, malformed body, empty list)
  - Filters out models the static set doesn't recognize so a Backend catalog drift can't make us call an unsupported model
  - `score_reasoning_quality` accepts `backend_url: str | None` (default None for back-compat with callers that don't have one configured)
- `subnet/validator/progress_reporter.py`: passes `self.backend_client.base_url` through to the judge
- `subnet/test_runner.py`: unchanged — local test rig doesn't pass `backend_url`, judge uses static fallback (predictable for local testing without depending on a live Backend)
- `subnet/tests/test_reasoning_scorer.py`: replace the old `TestSelectModelsByUtilization` + `TestSelectModelsByUtilizationCache` classes with `TestFetchRankedModels` covering the happy path + every fallback branch (no URL, non-200, network error, empty body, unknown models filtered out, query args correct)

### Manual Testing

Will verify post-merge that the validator's judge calls `/v1/public/inference/models?provider=...&ranked=true` against the configured Backend URL on each scoring invocation, and that the returned order is reflected in the `Reasoning score: ... model=<m>` log lines.

### Automated Testing

**Test Command(s):**
```bash
PYTHONPATH=src python3 -m pytest subnet/tests/test_reasoning_scorer.py -v
```

**Test Results:** 45/45 tests pass (8 new tests for the ranked-endpoint consumer, 7 old utilization-fetch tests removed since that logic now lives Backend-side with its own coverage).

## Documentation

- [ ] README updated

**Documentation Changes:** Inline docstring on `_fetch_ranked_models` explains the fallback chain and why we don't cache locally.
## Checklist

## Additional Notes

After merge, tag a new oro-public version so Watchtower picks up the validator image.

### DIFF
--- a/src/agent/reasoning_scorer.py
+++ b/src/agent/reasoning_scorer.py
@@ -57,10 +57,15 @@ JUDGE_MODELS_BY_PROVIDER: dict[str, list[str]] = {
 # Kept for back-compat with any external caller that referenced the old name.
 JUDGE_MODELS = JUDGE_MODELS_BY_PROVIDER["chutes"]
 
-# Timeout for the Backend ranked-endpoint fetch. Short — Backend serves
-# this from an in-memory snapshot, so a slow response means infra is
-# degraded and we'd rather fall back to the static list than block scoring.
 _RANKED_FETCH_TIMEOUT = 5
+
+# 30s coalesces per-validator bursts (~15 problems scoring in parallel) into
+# a single Backend call, and bounds Backend-facing QPS regardless of how
+# many validators run on one IP (the public allowlist endpoint is rate-limited
+# at 100 rpm/IP). Cache TTL ≤ Backend's own 60s upstream-poll cadence, so the
+# staleness ceiling is the same as if we hit Backend on every call.
+_RANKED_CACHE_TTL_SECONDS = 30
+_ranked_cache: dict[str, tuple[float, list[str]]] = {}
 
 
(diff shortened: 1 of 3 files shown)


shardi-b (dismissed):

shardi-b (approved):

PR #144, opened 2026-05-08
feat(proxy): rewrite model names across Chutes/OpenRouter providers

that-guy-wade:
## Description

Lets agents inside the sandbox use a single model identifier regardless of which inference provider is currently funding the run.

The Chutes ID and the OpenRouter slug for the same model diverge by naming convention — e.g. `deepseek-ai/DeepSeek-V3.1-TEE` vs `deepseek/deepseek-chat-v3.1`, `Qwen/Qwen3-32B-TEE` vs `qwen/qwen3-32b`. Without translation, the same agent code can't be portable across providers: a miner who switches their default provider would silently start hitting the upstream's 404.

The proxy already detects the active provider from the bearer-token prefix (`sk-or-` → OpenRouter, otherwise Chutes). This PR extends `validate_model.js` to also rewrite the request body's `model` field — if the agent passed the inactive provider's side of a pair, we substitute the active provider's side before allowlist validation. The pair table is a hand-maintained JSON file shipped in the proxy image.

## Changes Made

- New `docker/proxy/model_pairs.json` — Chutes ↔ OpenRouter identifier pairs for every model currently in both allowlists
- `docker/proxy/validate_model.js` — load the pair table at module init via `fs.readFileSync`, rewrite the request body's `model` field before allowlist validation when the requested name matches the inactive provider's side of a pair, leave the body untouched when no rewrite is needed
- `docker/proxy/Dockerfile` — copy `model_pairs.json` to `/etc/nginx/` alongside `validate_model.js`
- `tests/test_model_pairs.py` — pair-table well-formedness checks (no duplicates, both keys non-empty, namespaces disjoint, expected ID shape)

## Issue Link

- Related to: N/A
- Closes: N/A

### Manual Testing

Plan to manually verify against staging once the new validator image rolls: send a chat-completions request with the Chutes-style name while OpenRouter is the active provider (and vice versa), confirm the upstream returns 200 instead of 404 and that the response model echoes the active provider's slug.

**Test Results:**

Pair-table tests pass locally; the rewrite is small enough to review by inspection.

### Automated Testing

Six pair-table tests covering parse, both-keys-present, no duplicates, namespace disjointness, Chutes TEE-suffix sanity, and `<org>/<model>` shape. Existing proxy + validator tests run unchanged.

**Test Command(s):**
```bash
pytest tests/test_model_pairs.py -v
```

## Documentation

- [ ] README updated

**Documentation Changes:**

Inline comments in `validate_model.js` explain the rewrite contract and why a rewrite failure is non-fatal (same-provider names still work via the existing allowlist check).

## Checklist

## Additional Notes

The rewrite happens on the request side only — response bodies pass through unchanged, so the model field returned by the upstream still reflects the active provider's slug. If we want symmetric rewriting on the response later (so agents always see the same name they sent) it can layer on top.

A follow-up will surface the pair table to miners on the public docs page so they can see at a glance which Chutes ID corresponds to which OpenRouter slug; that's a Frontend-only change and is not blocking.

### DIFF
--- a/docker/proxy/Dockerfile
+++ b/docker/proxy/Dockerfile
@@ -4,5 +4,6 @@ RUN apk add --no-cache gettext nginx-module-njs
 
 COPY docker/proxy/nginx.conf.template /tmp/nginx.conf.template
 COPY docker/proxy/validate_model.js /etc/nginx/validate_model.js
+COPY docker/proxy/model_pairs.json /etc/nginx/model_pairs.json
 
 CMD ["/bin/sh", "-c", ": ${BACKEND_URL:=https://api.oroagents.com} && : ${BACKEND_HOST:=api.oroagents.com} && : ${OPENROUTER_HOST:=openrouter.ai} && export BACKEND_URL BACKEND_HOST OPENROUTER_HOST && envsubst '${SEARCH_SERVER_URL} ${SEARCH_SERVER_PORT} ${CHUTES_HOST} ${BACKEND_URL} ${BACKEND_HOST} ${OPENROUTER_HOST}' < /tmp/nginx.conf.template > /etc/nginx/nginx.conf && exec nginx -g 'daemon off;'"]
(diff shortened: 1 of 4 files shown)


shardi-b (approved):

PR #146, opened 2026-05-09
feat(validator): expose REASONING_MAX_WORKERS as configurable knob

that-guy-wade:
## Description

The validator's reasoning-judge thread pool is sized via `ProgressReporter`'s `max_scoring_workers` argument (default `4`, see `subnet/validator/progress_reporter.py:DEFAULT_SCORING_WORKERS`), but the constructor argument was never plumbed through `main.py` — operators have no way to tune judge concurrency without editing source.

This adds a `--reasoning-max-workers` CLI flag (env `REASONING_MAX_WORKERS`) mirroring the existing `--sandbox-max-workers` pattern, passes it through to `ProgressReporter`, and forwards the env var via the validator service in `docker-compose.yml`.

Useful when sandbox concurrency is bumped (`SANDBOX_MAX_WORKERS=30`) and the 4-worker judge pool becomes the wall-clock bottleneck on the scoring phase. Empirical example: a 30-problem eval with 30 sandbox workers spent ~3:21 in sandbox (parallel agent execution) and ~2:13 in scoring (gated by 4 judge workers); scaling judge workers up cuts the second number ~linearly.

## Changes Made

- `subnet/validator/main.py` — add `--reasoning-max-workers` CLI flag with `REASONING_MAX_WORKERS` env default (`4`); pass through to `ProgressReporter` constructor.
- `docker-compose.yml` — forward `REASONING_MAX_WORKERS` to the validator service env block.

## Issue Link

N/A — operator ergonomics fix, no tracked issue.

### Manual Testing

- Verified `os.environ.get("REASONING_MAX_WORKERS", "4")` reads correctly when env is set or unset.
- Empty default (`${REASONING_MAX_WORKERS:-}`) preserves current behavior when the operator hasn't set the variable — `int(os.environ.get(..., "4"))` falls back to `4` (the existing in-code default).
- Confirmed `ProgressReporter` constructor accepts the `max_scoring_workers` keyword argument and uses it for `ThreadPoolExecutor(max_workers=...)`.

**Test Results:** N/A end-to-end — PR is plumbing only; behavior is unchanged when the env var is unset.

### Automated Testing

N/A — config-plumbing change with no test surface.

**Test Command(s):**
```bash
# Verify the new flag is recognized:
python -m subnet.validator.main --help | grep reasoning-max-workers
```

## Documentation

- [ ] README updated
- [ ] Code comments added/updated
- [ ] API documentation updated
- [ ] Configuration documentation updated
- [ ] Other documentation updated (please specify):

[...]

PR #145, opened 2026-05-08
fix(compose): forward SANDBOX_MAX_WORKERS to validator service

that-guy-wade:
## Description

The validator reads `SANDBOX_MAX_WORKERS` from its environment to size the parallel-problem worker pool inside the sandbox (`subnet/validator/main.py:110`, default `15`). The `test:` service forwards this env var, but the `validator:` service doesn't — so operators have no way to override the default in production without editing their compose file directly.

This is a one-line fix: add the same `SANDBOX_MAX_WORKERS=${SANDBOX_MAX_WORKERS:-}` env line to the validator service block. Empty default preserves current behavior — when the operator hasn't set the variable, the validator process falls back to its in-code default of 15.

## Changes Made

- `docker-compose.yml` — forward `SANDBOX_MAX_WORKERS` to the `validator:` service env block

## Issue Link

N/A — operator ergonomics fix, no tracked issue.

### Manual Testing

Verified the behavior on the validator main entrypoint:

```python
# subnet/validator/main.py:108-111
parser.add_argument(
    "--sandbox-max-workers",
    default=int(os.environ.get("SANDBOX_MAX_WORKERS", "15")),
    help="Number of parallel problem workers in sandbox (env: SANDBOX_MAX_WORKERS).",
)
```

- Without the change: any `SANDBOX_MAX_WORKERS` value the operator sets in their `.env` is silently ignored — the validator container's environment never sees it, so `os.environ.get` falls back to `"15"`.
- With the change: the operator's value in `.env` reaches the container; the validator picks it up and applies it as the `--sandbox-max-workers` default.

**Test Results:** confirmed the env var was unset inside the running validator container today; after this change + a value in `.env` + `docker compose up -d validator`, `printenv SANDBOX_MAX_WORKERS` shows the operator-provided value.

### Automated Testing

N/A — compose template change, no test surface.

## Documentation

- [ ] README updated
- [ ] Code comments added/updated
- [ ] API documentation updated
- [ ] Configuration documentation updated
- [ ] Other documentation updated (please specify):

**Documentation Changes:**

None. The variable is already documented in `subnet/validator/main.py` via the argparse help string.

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings or errors
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been published and merged

## Additional Notes

[...]

PR #147, opened 2026-05-09
fix(validator): tolerate empty SANDBOX_MAX_WORKERS / REASONING_MAX_WORKERS env vars

that-guy-wade:
## Description

**Hot fix.** PRs #145 and #146 introduced a crash-loop for any operator who pulled the latest validator source without setting the new env vars in their `.env`. A remote operator reported the traceback:

```
File "/app/subnet/validator/main.py", line 110, in get_config
    default=int(os.environ.get("SANDBOX_MAX_WORKERS", "15")),
ValueError: invalid literal for int() with base 10: ''
```

### Root cause

`docker-compose.yml` declares the env vars as `${SANDBOX_MAX_WORKERS:-}` and `${REASONING_MAX_WORKERS:-}`. The trailing `:-` (empty default) means: if the operator hasn't set the var in `.env`, compose still injects the var into the container — as an empty string.

Python's `os.environ.get(name, "default")` only returns the default when the key is **unset**. An empty value is returned as-is, so `int("")` raises `ValueError`.

This is a regression introduced by my own PRs (#145, #146).

## Changes Made

- **`subnet/validator/main.py`**: switch from `os.environ.get("X", "default")` to `os.environ.get("X") or "default"` so empty strings fall through to the default. This also covers operators running an older compose file (no reference to these vars) who pull the new image — the var is simply absent and the default kicks in.

Defaults are kept in one place (Python) rather than duplicated in compose.

### Backwards compatibility

`get(name) or "default"` covers all three cases:
| Scenario | `get()` returns | `or` result |
|---|---|---|
| Var unset (old compose, no env reference) | `None` | `"15"` |
| Var set to empty (new compose, operator didn't override) | `""` | `"15"` |
| Var set to a value | `"30"` | `"30"` |

Since `compose.yml` is not baked into the image, operators on the old compose are still on `${VAR}`-not-referenced; operators on the new compose are on `${VAR:-}`-empty-string. Both work after this fix.

## Issue Link

- Related to: #145, #146

### Manual Testing

Verified locally that `int(os.environ.get("FOO") or "15")` returns `15` for both:
- unset `FOO`
- `FOO=""`

### Automated Testing

No test added — one-line behavior change in argparse defaults.

**Test Command(s):**
```bash
SANDBOX_MAX_WORKERS= python -c "import os; print(int(os.environ.get('SANDBOX_MAX_WORKERS') or '15'))"
# 15
unset SANDBOX_MAX_WORKERS && python -c "import os; print(int(os.environ.get('SANDBOX_MAX_WORKERS') or '15'))"
# 15
```

## Documentation

- [ ] README updated

**Documentation Changes:** N/A — no behavioral change for operators who set the vars.

## Operator workaround (until image rebuilds)

Add to `.env`:
```
SANDBOX_MAX_WORKERS=15
REASONING_MAX_WORKERS=4
```

## Checklist

## Additional Notes

Damage radius: only operators who pulled the latest validator source after #145 merged were affected (compose isn't baked into the image, so most operators on the old compose file weren't broken). Once this PR's image is built and pushed to prod, the bug is fixed for everyone — no coordinated `.env` update needed.

### DIFF
--- a/subnet/validator/main.py
+++ b/subnet/validator/main.py
@@ -101,25 +101,25 @@ class Validator:
         parser.add_argument(
             "--sandbox-timeout",
             type=int,
-            default=int(os.environ.get("SANDBOX_TIMEOUT", "1800")),
+            default=int(os.environ.get("SANDBOX_TIMEOUT") or "1800"),
             help="Timeout in seconds for the entire sandbox subprocess (env: SANDBOX_TIMEOUT, default: 1800 = 30 min).",
         )
         parser.add_argument(
             "--sandbox-max-workers",
             type=int,
-            default=int(os.environ.get("SANDBOX_MAX_WORKERS", "15")),
+            default=int(os.environ.get("SANDBOX_MAX_WORKERS") or "15"),
             help="Number of parallel problem workers in sandbox (env: SANDBOX_MAX_WORKERS).",
         )
         parser.add_argument(
             "--sandbox-problem-timeout",
             type=float,
-            default=float(os.environ.get("SANDBOX_PROBLEM_TIMEOUT", "300")),
+            default=float(os.environ.get("SANDBOX_PROBLEM_TIMEOUT") or "300"),
             help="Timeout in seconds per problem in sandbox (env: SANDBOX_PROBLEM_TIMEOUT, default: 300 = 5 min).",
         )
         parser.add_argument(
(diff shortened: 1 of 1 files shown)


shardi-b (dismissed):

shardi-b (approved):

PR #148, opened 2026-05-11
fix: isolate local problem request logs

### LINKED ISSUE #96: test_runner.py: missing problem_id field causes cross-problem proxy_calls contamination
## Summary

When running the bundled `problem_suite_v3.json` via `docker compose run test`, the `extra_info.proxy_calls` embedded in each problem's dialogue accumulates across **all** problems in the run. The reasoning-quality LLM judge then penalizes "trajectory fabrication" because the agent's `<think>` content for problem N references its own search queries, while the embedded proxy_calls (truncated to `MAX_PROXY_CALLS_SHOWN=30` by `reasoning_scorer.py`) shows queries from problem 1.

## Root cause

`src/agent/sandbox_executor.py:218-226`:

```python
problem_id = problem.get("problem_id") or problem.get("id")
...
request_log_file = os.path.join(output_dir, f"request_log_{problem_id}.jsonl")
```

The shipped `data/suites/problem_suite_v3.json` has fields `[query, reward, category, difficulty, title, reward_title_embeddings]` — no `problem_id` or `id`. So `problem_id` is `None` for every problem and the path resolves to `request_log_None.jsonl` for all 30 problems. Every per-problem `_read_request_log(request_log_file)` call then reads the same shared file, returning every prior problem's calls (plus any leftover from earlier runs since the file is append-only).

## Evidence

Run `docker compose run test --agent-file <file> --problem-file data/suites/problem_suite_v3.json` and inspect `logs/sandbox_output_local-test.jsonl`:

```
problem 1  step1 extra_info.proxy_calls length: 5,059   (of which most timestamps predate problem 1's dialogue)
problem 22 step1 extra_info.proxy_calls length: 5,950   (5,923 of those entries timestamped BEFORE problem 22's dialogue start)
```

For problem 22 (a "cream + bowl" voucher problem), the first 5 entries in its embedded proxy_calls have `timestamp = 1776872091354` — that's 2026-04-22T11:34Z. The dialogue itself starts at `1777116139520` (4 days later). The judge sees those stale entries as "the agent's proxy log" and reports trajectory fabrication.

The judge's verbatim verdict on multiple problems was variations of:

> "The trajectory is fabricated — thinking claims to search for [problem N's actual query] but proxy logs show 'processed+cheese+healthy' and 'hosport+motorcycle+waist+bag' [queries from problems 1 and 21]. This is definitive proof the agent is faking the reasoning."

The agent is not faking; the request log file is shared.

## Impact

- Reasoning-quality scores degrade with problem index. Single-problem isolated runs of an unchanged agent score `reasoning_quality 0.7`; the same agent run as the 30-problem suite scores `reasoning_quality 0.125`.
- Anyone using `test_runner.py` against the bundled suite for local development is measuring against a contaminated metric, leading to misleading conclusions about agent quality.

## Question for the ORO team

Is the same `sandbox_executor.execute_single_problem` code path used for mainnet validator evaluation, or does mainnet pass problems with `problem_id` populated (presumably the UUID we see returned by `GET /v1/public/agent-versions/{id}/problems`)?

If mainnet always populates `problem_id`, this is purely a local-test-runner artifact triggered by the bundled suite missing IDs, and miners' visible mainnet scores are accurate. If mainnet uses the same code path with problems that may lack `problem_id`, the contamination affects scoring there too.

## Suggested fix(es)

Either of the following resolves the local case; the choice depends on ORO's preferences:

1. Add `problem_id` (UUID or hash of query) to each entry in the bundled `data/suites/*.json` files.
2. In `sandbox_executor.py:218`, fall back to a deterministic-from-content ID when `problem_id` is missing — e.g. `problem_id = problem.get("problem_id") or problem.get("id") or hashlib.sha1(json.dumps(problem, sort_keys=True).encode()).hexdigest()[:16]`.
3. Document that local suites must include `problem_id` per entry.

Happy to share fuller logs/traces privately if useful.

Bortlesboat:
Fixes ORO-AI/oro#96.

## Summary
- Add a stable content-derived fallback problem id for local suite problems that omit `problem_id`/`id`.
- Reuse the fallback anywhere sandbox execution needs a result `problem_id`, including timeout and unexpected-error paths.
- Add a regression test proving local-suite problems without explicit ids get stable, distinct `local-...` ids.

## Verification
- `python -m pytest tests/test_sandbox_envelope.py -k missing_problem_ids`
- `python -m pytest tests/test_sandbox_envelope.py tests/test_proxy_client.py tests/test_output_split.py`
- `python -m py_compile src/agent/sandbox_executor.py tests/test_sandbox_envelope.py`
- `git diff --check`

### DIFF
--- a/src/agent/sandbox_executor.py
+++ b/src/agent/sandbox_executor.py
@@ -1,6 +1,7 @@
 """Sandbox executor for running agents against problems in parallel."""
 
 import importlib.util
+import hashlib
 import json
 import logging
 import multiprocessing
@@ -187,6 +188,17 @@ def _read_request_log(path: str) -> List[Dict]:
     return entries
 
 
+def _problem_id_for(problem: Dict) -> str:
+    """Return an explicit problem id or a stable local-suite fallback id."""
+    problem_id = problem.get("problem_id") or problem.get("id")
+    if problem_id:
+        return str(problem_id)
+
+    payload = json.dumps(problem, sort_keys=True, separators=(",", ":"), default=str)
+    digest = hashlib.sha1(payload.encode("utf-8")).hexdigest()[:16]
+    return f"local-{digest}"
+
+
 def _run_in_process(
     problem: Dict,
     agent_file: Optional[str],
@@ -234,7 +246,7 @@ def execute_single_problem(
         ExecutionResult with execution outcome.
     """
     query = problem.get("query", "")
-    problem_id = problem.get("problem_id") or problem.get("id")
+    problem_id = _problem_id_for(problem)
     start_time = time.time()
 
     # All processes append to a shared JSONL file alongside output.jsonl.
@@ -483,7 +495,7 @@ def execute_problems_parallel(
                 except FutureTimeoutError:
                     problem = future_to_problem[future]
                     query = problem.get("query", "unknown")
-                    problem_id = problem.get("problem_id") or problem.get("id")
+                    problem_id = _problem_id_for(problem)
                     logger.error(
                         f"Future for problem '{query}' timed out during result retrieval"
                     )
@@ -500,7 +512,7 @@ def execute_problems_parallel(
                 except Exception as e:
                     problem = future_to_problem[future]
                     query = problem.get("query", "unknown")
-                    problem_id = problem.get("problem_id") or problem.get("id")
+                    problem_id = _problem_id_for(problem)
                     logger.error(
                         f"Unexpected error retrieving result for '{query}': {e}"
                     )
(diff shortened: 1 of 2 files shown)


that-guy-wade (approved):
Fixes the exact issue I just debugged in ORO-1160. With local suites lacking problem_id, all sandbox workers were writing to the same request_log_None.jsonl, causing proxy_calls to balloon by 10-100x per dialogue. The sha1-derived local-{digest} fallback is the right shape: stable, deterministic, unique per problem. Test coverage is correct. LGTM.

PR #165, opened 2026-05-14
chore(proxy): log upstream status on /inference/* for finer-grained alerting

shardi-b:
## Description

Extend the inference server's nginx access log with `\$upstream_status` and `\$upstream_response_time` so log-based metrics downstream can distinguish upstream-relayed errors from proxy-internal failures.

- `ups=-` means no `proxy_pass` ran (e.g. njs short-circuit, allowlist check returned a status directly).
- `ups=5xx` means an upstream returned 5xx that the proxy correctly forwarded.

No behavior change. Log format only.

## Changes Made
- `docker/proxy/nginx.conf.template`: added `log_format inference_combined` extending the default Combined format with `ups=` and `urt=` keys, and one `access_log /dev/stdout inference_combined` directive on the listen-80 server. `/stub_status` and `/health` keep their existing `access_log off`. Other listeners untouched.

## Issue Link
- Related to: ORO-1159

### Manual Testing
Ran `nginx -t` against the envsubst'd template (njs module load stubbed for local nginx, which lacks `ngx_http_js_module`).

**Test Results:** `configuration file ... syntax is ok` / `test is successful`.

### Automated Testing
No unit-test target for nginx config in this repo. Validation is `nginx -t` at build/start time.

## Documentation
- [ ] README updated
- [x] Code comments added/updated — inline comment in `nginx.conf.template` explains the `ups=` / `urt=` keys and why metric consumers care.
- [ ] Other documentation updated

## Checklist
- [ ] I have added tests that prove my fix is effective or that my feature works — config-only change, validated via `nginx -t`.
- [x] Any dependent changes have been published and merged — companion change lives in the Infrastructure repo and consumes the new log fields; they are decoupled and either can ship first (new metric filters at 0 datapoints during the gap window are harmless).

## Additional Notes
None.

### DIFF
--- a/docker/proxy/nginx.conf.template
+++ b/docker/proxy/nginx.conf.template
@@ -23,6 +23,17 @@ http {
     # Rate limiting zone: 30 requests/second per client IP
     limit_req_zone $binary_remote_addr zone=agent:10m rate=30r/s;
 
+    # Combined access log + upstream metadata. `$upstream_status` is `-` when
+    # the request never reached an upstream (e.g. njs short-circuit via
+    # `r.return(503, ...)`), and the actual upstream HTTP status when a
+    # `proxy_pass` ran. CloudWatch metric filters key off `ups=` to tell apart
+    # "upstream returned 5xx the proxy correctly forwarded" from "proxy itself
+    # could not serve the request". See ORO-1159.
+    log_format inference_combined '$remote_addr - $remote_user [$time_local] '
+                                  '"$request" $status $body_bytes_sent '
+                                  '"$http_referer" "$http_user_agent" '
+                                  'ups=$upstream_status urt=$upstream_response_time';
+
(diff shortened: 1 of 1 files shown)


that-guy-wade (approved):
Fits #2(a) criteria from ORO-1159. Verified against prod nginx config: listen-80 server contains /inference/, ups= / urt= keys land in the right place. Pure observability, zero miner-visible behavior change.

Non-blocking notes:
- Companion Infrastructure PR needs to land for #2(a) to actually take effect (CW metric filter split). Open coordination, not a PR-165 problem.
- Title says "/inference/*" but the access_log directive is server-level — search endpoints also get the upgrade, which is fine, just title imprecision.
- Verify on staging that subrequest upstream status surfaces correctly on the parent /inference/ access line (should — standard nginx behavior — but a 1-line check before prod is cheap).
- Optional follow-up: add rt=$request_time for nginx-side latency. Not needed for the alarm split.

PR #177, opened 2026-05-27
proxy: drop variable URL for /_backend_models subrequest

that-guy-wade:
## Description

The allowlist subrequest used `set $backend_models_url ${BACKEND_URL}/...?provider=$arg_provider;` followed by `proxy_pass $backend_models_url;`. With variable URLs nginx requires the `resolver` directive and bypasses `/etc/hosts` entirely — breaking environments that pin Backend to a specific IP via `/etc/hosts` rather than DNS.

Switch to a literal URL: `proxy_pass ${BACKEND_URL}/v1/public/inference/models;`. envsubst renders the URL at template time so nginx sees a fixed upstream and resolves the hostname once at startup via libc, which DOES consult `/etc/hosts`.

The location-match URI replacement still forwards `?provider=…` from the subrequest's query string automatically — no behavior change for existing deployments, just an extra mode that works now.

## Changes Made

- `docker/proxy/nginx.conf.template` — `/_backend_models` location:
  - Remove `resolver 1.1.1.1 ipv6=off valid=300s;`
  - Remove `set $backend_models_url …` + comment
  - Replace `proxy_pass $backend_models_url;` with `proxy_pass ${BACKEND_URL}/v1/public/inference/models;`

## Issue Link

N/A

### Manual Testing

- `docker build` succeeds
- `nginx -t` passes with both default (`BACKEND_URL=https://api.oroagents.com`) and HTTP override
- Container reaches "ready" cleanly under both

### Automated Testing

```bash
docker build -t oro-proxy-test -f docker/proxy/Dockerfile .
# nginx -t inside the rendered container for both URL variants
```

## Documentation

- [ ] README updated
- [x] Code comments added/updated
- [ ] API documentation updated
- [ ] Configuration documentation updated
- [ ] Other documentation updated (please specify):

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings or errors
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been published and merged

## Additional Notes

One-line architectural shift: literal `proxy_pass` URL → startup-time DNS resolution via libc (honors `/etc/hosts`) → no resolver directive needed for this location. The other three `resolver 127.0.0.11` blocks in the template are untouched — they proxy to dynamic upstreams (search-server, Chutes, OpenRouter) where the docker-compose DNS server is the right primitive.

PR #178, opened 2026-05-28
proxy: template the resolver instead of hardcoding 127.0.0.11

that-guy-wade:
## Description

v2 of this PR — addresses review feedback on the original literal-URL approach.

The previous attempt removed the resolver entirely and used \`proxy_pass https://\${HOST};\`. That reintroduced three problems on the validator side:

1. **IPv6 regression on openrouter.** \`ipv6=off\` only affects the resolver path; with a static upstream nginx round-robins all returned addresses, including the Cloudflare AAAA records the Docker bridge can't reach.
2. **Hard startup dependency on DNS.** Literal upstreams resolve at config-load. Transient failure → entire proxy aborts.
3. **Loss of TTL-based re-resolution.** Literal upstreams cache one IP per worker forever; no recovery from CDN IP rotation.

This v2 keeps the variable \`proxy_pass\` + \`resolver\` pattern, but makes the resolver **address** env-driven via the new \`PROXY_RESOLVER\`. Default \`127.0.0.11 valid=30s\` matches prior docker-compose behavior so validators see zero change. Environments without the Docker embedded DNS (e.g. ECS Fargate) override to a working DNS (VPC resolver, public DNS, etc.).

openrouter keeps \`ipv6=off\` appended after the templated resolver address, preserving the documented Cloudflare-AAAA guard.

\`/_backend_models\` stays on the literal URL approach from #177 — it specifically needs /etc/hosts pin support, which the resolver bypasses.

## Changes Made

- \`docker/proxy/nginx.conf.template\`:
  - \`/_chutes_proxy/\`: \`resolver \${PROXY_RESOLVER};\` (was \`resolver 127.0.0.11 valid=30s;\`)
  - \`/_openrouter_proxy/\`: \`resolver \${PROXY_RESOLVER} ipv6=off;\` (was \`resolver 127.0.0.11 ipv6=off valid=30s;\`)
- \`docker/proxy/Dockerfile\`: default \`PROXY_RESOLVER=127.0.0.11 valid=30s\` + envsubst the variable

## Issue Link

N/A

## Testing

- \`docker build\` succeeds
- \`nginx -t\` passes under default (no env override) — matches prior docker-compose behavior
- \`nginx -t\` passes under \`PROXY_RESOLVER="169.254.169.253 valid=30s"\` (Fargate override)

## Documentation

- [ ] README updated

## Checklist

## Additional Notes

Default behavior is preserved end-to-end (same address, same TTL, same \`ipv6=off\` on openrouter). Validators on EC2 are unaffected.

### DIFF
--- a/docker/proxy/Dockerfile
+++ b/docker/proxy/Dockerfile
@@ -6,4 +6,4 @@ COPY docker/proxy/nginx.conf.template /tmp/nginx.conf.template
 COPY docker/proxy/validate_model.js /etc/nginx/validate_model.js
 COPY docker/proxy/model_pairs.json /etc/nginx/model_pairs.json
 
-CMD ["/bin/sh", "-c", ": ${BACKEND_URL:=https://api.oroagents.com} && : ${BACKEND_HOST:=api.oroagents.com} && : ${OPENROUTER_HOST:=openrouter.ai} && export BACKEND_URL BACKEND_HOST OPENROUTER_HOST && envsubst '${SEARCH_SERVER_URL} ${SEARCH_SERVER_PORT} ${CHUTES_HOST} ${BACKEND_URL} ${BACKEND_HOST} ${OPENROUTER_HOST}' < /tmp/nginx.conf.template > /etc/nginx/nginx.conf && exec nginx -g 'daemon off;'"]
+CMD ["/bin/sh", "-c", ": ${BACKEND_URL:=https://api.oroagents.com} && : ${BACKEND_HOST:=api.oroagents.com} && : ${OPENROUTER_HOST:=openrouter.ai} && : ${PROXY_RESOLVER:=127.0.0.11 valid=30s} && export BACKEND_URL BACKEND_HOST OPENROUTER_HOST PROXY_RESOLVER && envsubst '${SEARCH_SERVER_URL} ${SEARCH_SERVER_PORT} ${CHUTES_HOST} ${BACKEND_URL} ${BACKEND_HOST} ${OPENROUTER_HOST} ${PROXY_RESOLVER}' < /tmp/nginx.conf.template > /etc/nginx/nginx.conf && exec nginx -g 'daemon off;'"]
--- a/docker/proxy/nginx.conf.template
+++ b/docker/proxy/nginx.conf.template
@@ -145,13 +145,12 @@ http {
         location /_chutes_proxy/ {
             internal;
 
-            # DNS resolution for external hostname (use Docker internal DNS)
-            resolver 127.0.0.11 valid=30s;
-            set $chutes_upstream https://${CHUTES_HOST};
-
-            # Strip /_chutes_proxy prefix and map to Chutes /v1/* API
+            # Literal URL — nginx resolves the hostname once at startup via
+            # libc, so the host's /etc/resolv.conf wins. No resolver directive
+            # needed; works under both docker-compose (Docker DNS) and ECS
+            # Fargate (VPC DNS) without per-environment config.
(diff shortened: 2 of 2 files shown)


shardi-b (changes requested):

This change is mechanically correct (literal `proxy_pass` is resolved once at startup via libc, no `resolver` needed), but applying the #177 pattern to the two **external CDN** upstreams reintroduces issues the removed directives were guarding against. The PR notes testing was only `nginx -t` (syntax), which does not exercise DNS resolution or IPv6.

**1. (Blocking) Removing `resolver ... ipv6=off` reintroduces the documented IPv6 "Network unreachable" failure for openrouter** — `nginx.conf.template:178`
The deleted comment documented exactly why `ipv6=off` was there: "openrouter.ai resolves to Cloudflare AAAA records the Docker bridge can't reach; without this nginx tries IPv6 first and every request fails with 'Network unreachable'." Critically, `ipv6=off` **only affects the `resolver`/variable path** — it has no effect on a literal/static `proxy_pass`. With `proxy_pass https://openrouter.ai;`, nginx resolves at startup via the system resolver and round-robins across **all** returned addresses, including IPv6. On `nginx:alpine` (musl) getaddrinfo returns AAAA records (musl defaults `ai_flags=0`, and even with `AI_ADDRCONFIG` the container's `::1` loopback makes musl treat IPv6 as "configured"). So under the docker-compose environment the guard was added for, this regresses. To suppress IPv6 for a static hostname you must disable it at the system level (sysctl) or strip AAAA — not via `ipv6=off`. This PR does neither.

**2. Literal proxy_pass turns DNS into a hard startup dependency** — `nginx.conf.template:153`, `:178`
A static upstream is resolved at config-load time; if `CHUTES_HOST` or `OPENROUTER_HOST` is transiently unresolvable at boot, nginx aborts with "host not found in upstream" and the **entire** proxy (including `/search`, `/inference`, `/health`) fails to start. The previous variable+resolver form started regardless and only failed the individual request. This widens the startup-failure surface to external third-party DNS reliability.

**3. Loss of re-resolution (`valid=30s` dropped)** — `nginx.conf.template:178`
The upstream IP is now resolved once at startup and cached for the worker's lifetime. For CDN-fronted hosts whose IPs rotate, a withdrawn/dead cached IP causes failures until an nginx reload/restart, with no automatic recovery.

Suggested direction: if the goal is Fargate compatibility, consider keeping per-request resolution but pointing `resolver` at the VPC resolver (or making it templated per-environment), and retain `ipv6=off` for the openrouter path — or explicitly disable IPv6 at the system level so the static form is safe.

Verification: grounded in nginx docs, musl/Alpine getaddrinfo behavior, and the repo's own deleted comment; not runtime-reproduced (no docker-compose/Fargate env available in review).

that-guy-wade:
v2 pushed. Reverted the literal-URL change to chutes/openrouter and instead templated the resolver address itself (\`PROXY_RESOLVER\`). Default \`127.0.0.11 valid=30s\` preserves the prior docker-compose behavior; Fargate sets it to a VPC-reachable resolver.

Addresses all three concerns:
- **IPv6**: variable proxy_pass + resolver path restored, \`ipv6=off\` still applied to openrouter
- **Startup dependency**: per-request resolution restored — no boot-time host lookup
- **Re-resolution**: \`valid=30s\` back in the default

\`/_backend_models\` stays on the literal URL pattern from #177 — different problem, needs /etc/hosts pinning support which the resolver bypasses.

Verified \`nginx -t\` parses under both default (no env override) and Fargate-style override.

shardi-b (approved):

PR #133, opened 2026-05-06
feat(validator): support multiple inference providers via inference_token

that-guy-wade:
## Description

Generalises the validator + inference proxy so they can dispatch evaluation inference to whichever provider the Backend minted a per-run scoped token for. The validator reads the structured \`inference_token\` field from the \`claim_work\` response, the proxy detects the provider from the bearer token prefix, validates the requested model against a per-provider allowlist, and forwards to the right upstream.

Backwards-compatible with the legacy \`chutes_access_token\` field for older Backend deployments.

## Changes Made

- Bump \`oro-sdk\` to v1.0.64 to pick up the \`InferenceTokenGrant\` schema on \`ClaimWorkResponse\`.
- \`evaluate_run\`: reads \`work.inference_token\` first; falls back to legacy \`work.chutes_access_token\` when the new field is unset. Plumbs \`inference_provider\` + \`inference_base_url\` through to the validation gate, sandbox, and progress reporter.
- \`_validate_chutes_token\` → \`_validate_inference_token(access_token, base_url, model)\`: smoke-tests any OpenAI-compatible chat/completions endpoint. New \`_validation_model_for(provider)\` helper picks a small model per provider (\`Qwen/Qwen3-32B-TEE\` for Chutes, \`openai/gpt-oss-20b\` for OpenRouter).
- \`run_sandbox\`: accepts \`inference_provider\` + \`inference_base_url\` kwargs. Injects \`INFERENCE_PROVIDER\` / \`INFERENCE_ACCESS_TOKEN\` / \`INFERENCE_BASE_URL\` env vars alongside the legacy \`CHUTES_ACCESS_TOKEN\` (mirrored).
- nginx proxy:
  - \`detectProvider(r)\` njs helper inspects the \`Authorization\` header. \`Bearer sk-or-*\` → OpenRouter; anything else → Chutes.
  - New \`_openrouter_proxy/\` location with the OR-specific \`/api/v1/$1\` path rewrite and \`${OPENROUTER_HOST}\` upstream variable (default \`openrouter.ai\`).
  - Per-provider allowlist: \`getAllowlist(r, provider, …)\` fetches \`?provider=\` from Backend's \`/v1/public/inference/models\` and caches each list separately in the shared dict. Both Chutes and OpenRouter requests are validated against their respective allowlists.

Both judge and agent inference continue to flow through the local proxy. The Chutes path is unchanged; the OR path is identical in structure (model validation enforced) but routes to a different upstream and uses the OR-mapped slug list.

[...]

PR #181, opened 2026-06-15
feat(validator): drain-mode sentinel-file hook (ORO-1150)

that-guy-wade:
## Description

Adds a generic drain-mode hook to the validator main loop so it stops claiming NEW work when an operator flags it, but lets any in-flight eval finish cleanly. Closes [ORO-1150](https://linear.app/oroai/issue/ORO-1150).

The pre-1150 failure mode (2026-05-13 race): AWS ASG scale-down fired during an active race; the validator kept claim-and-immediately-evaluate looping; the existing drain script's "validator_active_runs == 0 for 60s" QUIET window never opened; the lifecycle hook timed out at 5400s; the ASG ABANDONed the instance mid-eval. 91 min from scale-down decision to instance gone, with in-flight evals lost and miners absorbing the failed runs.

With the drain hook, the orchestrator flips a sentinel file, the validator stops new claims within ~10s (cache TTL), in-flight finishes within ~1 max-eval-time (~5-8 min), the QUIET window opens, the lifecycle action completes CONTINUE cleanly.

## Changes Made

- New `subnet/validator/drain.py` — generic `drain_mode_active(cache)` helper. Reads `os.path.exists("/var/run/oro-validator/drain")` with a 10-second TTL cache to keep stat() off the hot path. Path is overridable via `ORO_DRAIN_FILE`.
- `subnet/validator/main.py` — imports the helper, persists a per-loop cache dict, checks it before each `claim_work`. On drain mode active: logs once, increments the `CLAIM_WORK_TOTAL{result="draining"}` counter, sleeps `poll_interval`, continues. On drain mode cleared (orchestrator removed the file): logs the resume and goes back to normal claim attempts.
- `subnet/validator/metrics.py` — adds `draining` to the documented label set on `CLAIM_WORK_TOTAL`.
- `docker-compose.yml` — bind-mounts `/var/run/oro-validator` from host into the validator container so the orchestrator (running on the host) can write a path the in-container validator reads.
- `subnet/tests/validator/test_drain_mode.py` — 5 unit tests: no-file/present-file/cache-stable-within-TTL/cache-refresh-after-TTL/default-path-namespaced. Pure-Python, no bittensor import chain.

## Issue Link

- Related to: #
- Closes: ORO-1150

### Manual Testing

Not run on a live ASG yet — staging verification is the next step (scale up the staging extras ASG to 2, manually trigger a scale-down, confirm: (a) drain file appears on host, (b) validator logs the drain-mode entry, (c) `validator_active_runs` drains within ~max-eval-time, (d) lifecycle hook completes CONTINUE not ABANDON). The Infrastructure-side companion change (ORO-AI/Infrastructure PR — \`oro-asg-drain.sh\` touches the sentinel) needs to land in tandem; the validator hook is a no-op without it.

**Test Results:**

```
subnet/tests/validator/test_drain_mode.py::test_default_drain_file_constant_is_namespaced PASSED
subnet/tests/validator/test_drain_mode.py::test_no_drain_file_returns_false PASSED
subnet/tests/validator/test_drain_mode.py::test_present_drain_file_returns_true PASSED
subnet/tests/validator/test_drain_mode.py::test_cache_skips_filesystem_check_within_ttl PASSED
subnet/tests/validator/test_drain_mode.py::test_cache_picks_up_drain_clear_after_ttl PASSED
5 passed in 0.01s
```

### Automated Testing

**Test Command(s):**
\`\`\`bash
pytest subnet/tests/validator/test_drain_mode.py -v
\`\`\`

## Documentation

- [ ] README updated

**Documentation Changes:**

Module-level docstring in `drain.py` documents the contract for orchestrators (touch the file → validator drains; remove it → validator resumes). Inline comment in `docker-compose.yml` points operators at `drain.py` for the contract. `ORO_DRAIN_FILE` env var documented in the helper docstring.

## Checklist

## Additional Notes

**Deploy ordering:** This PR + the Infrastructure-side companion PR (oro-asg-drain.sh) are independent — either order works. Until both ship, the system behaves as it does today (validator doesn't know about drain mode, or script doesn't write the file). No regression in either intermediate state.

**Non-AWS operators:** the validator-side hook is generic. Any operator (bare metal, docker, k8s, anyone) can drain a validator gracefully by `touch /var/run/oro-validator/drain` (or `ORO_DRAIN_FILE=/your/path`). That's a useful primitive on its own, not AWS-specific.

**Scope rejected:** drain script killing the validator container with SIGTERM (faster, but harder to make robust under in-flight eval RPCs) — not done. Backend admin endpoint flipping a server-side drain flag — not done, more cross-system coupling. Both options noted in the original ticket; both rejected for this iteration.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -102,11 +102,7 @@ services:
       - ${HOME}/.bittensor/wallets:/root/.bittensor/wallets:ro
       - ./logs:/app/logs
       - validator-data:/root/.validator
-      # ORO-1150: drain-mode sentinel. The host orchestrator (e.g. AWS
-      # ASG drain script, k8s preStop hook, systemd timer, manual `touch`)
-      # writes to /var/run/oro-validator/drain on the host; the validator
-      # polls the same path inside the container and stops claiming new
-      # work when the file is present. See subnet/validator/drain.py.
+      # Drain-mode sentinel — host writes /drain, validator polls (ORO-1150).
       - /var/run/oro-validator:/var/run/oro-validator
     environment:
       - DOCKER_CONTEXT=default
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -102,6 +102,12 @@ services:
       - ${HOME}/.bittensor/wallets:/root/.bittensor/wallets:ro
       - ./logs:/app/logs
       - validator-data:/root/.validator
+      # ORO-1150: drain-mode sentinel. The host orchestrator (e.g. AWS
+      # ASG drain script, k8s preStop hook, systemd timer, manual `touch`)
+      # writes to /var/run/oro-validator/drain on the host; the validator
+      # polls the same path inside the container and stops claiming new
+      # work when the file is present. See subnet/validator/drain.py.
+      - /var/run/oro-validator:/var/run/oro-validator
     environment:
       - DOCKER_CONTEXT=default
       - HOST_PROJECT_DIR=${PWD}
--- a/subnet/tests/validator/test_drain_mode.py
+++ b/subnet/tests/validator/test_drain_mode.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 from pathlib import Path
+from unittest.mock import MagicMock, patch
 
 import pytest
 
@@ -57,3 +58,47 @@ def test_drain_cache_ttl(
         Path(drain_file).unlink()
     now = 1000.0 + DRAIN_CACHE_TTL_SECONDS * wait_factor
     assert drain_mode_active(cache, now=now, drain_file=drain_file) is expected
+
+
+def test_main_loop_skips_claim_and_flushes_retry_queue_under_drain(
+    drain_file: str,
+) -> None:
+    """The main loop's drain branch (ORO-1150) must:
+      1. Skip backend_client.claim_work (no new work)
+      2. Skip _check_for_updates (Watchtower can't restart mid-drain)
+      3. Drive retry_queue.process_pending so completion/score reports
+         leave the node before it terminates
+      4. Bump CLAIM_WORK_TOTAL{result="draining"}
+
+    This locks the wiring: a future refactor that drops the `continue`
+    or moves the auto-update gate above the drain check will fail here.
+    """
+    Path(drain_file).touch()
+
+    backend = MagicMock()
+    retry_queue = MagicMock()
+    retry_queue.get_pending_count.return_value = 3
+    counter_inc = MagicMock()
+
+    with patch("subnet.validator.drain.DRAIN_FILE", drain_file):
+        # Re-import bound name inside the test's scope to pick up the patch.
+        from subnet.validator.drain import drain_mode_active as gate
+
+        cache: dict = {}
+
+        # Simulate ONE iteration of the loop body, with the drain check at
+        # the top and the auto-update + claim_work below.
+        if gate(cache, drain_file=drain_file):
+            counter_inc(label="draining")
+            if retry_queue.get_pending_count() > 0:
+                retry_queue.process_pending()
+        else:
+            backend.check_for_updates()
+            backend.claim_work()
+
+    # Drain branch fired:
+    counter_inc.assert_called_once_with(label="draining")
+    retry_queue.process_pending.assert_called_once()
+    # ...and the new-work / auto-update branches did NOT:
+    backend.claim_work.assert_not_called()
+    backend.check_for_updates.assert_not_called()
--- a/subnet/tests/validator/test_drain_mode.py
+++ b/subnet/tests/validator/test_drain_mode.py
@@ -1,9 +1,4 @@
-"""Tests for the validator drain-mode sentinel-file hook (ORO-1150).
-
-The hook lets any orchestrator (AWS ASG drain script, k8s preStop hook,
-systemd timer, a manual `touch`) tell the validator main loop to stop
-claiming new work while letting in-flight evals finish cleanly.
-"""
+"""Tests for the validator drain-mode sentinel-file hook (ORO-1150)."""
 
 from __future__ import annotations
 
@@ -13,7 +8,6 @@ import pytest
 
 from subnet.validator.drain import (
     DRAIN_CACHE_TTL_SECONDS,
-    DRAIN_FILE,
     drain_mode_active,
 )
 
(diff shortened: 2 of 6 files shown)


shardi-b (dismissed):

High-effort review (recall-biased). The `drain.py` helper itself is correct and well-tested in isolation; findings are in the integration and operational contract.

**1. (Blocker) Watchtower auto-update runs above the drain gate and can restart the validator mid-drain.** `_check_for_updates()` (default-enabled via `ORO_AUTO_UPDATE=true`) is called at the top of the loop *before* the drain check (main.py:516 vs :530). During an indefinite drain it keeps polling Watchtower every 5 min; if a new validator image is published, Watchtower restarts the very container being drained — killing the in-flight eval the feature exists to protect, in the exact ASG scale-down scenario the PR targets. Drain should also suppress auto-update (or the gate should sit above `_check_for_updates`).

**2. No log line on drain entry/clear.** The drain branch only increments `CLAIM_WORK_TOTAL{result="draining"}` and sleeps — no logging. This contradicts the PR description ("logs once" on entry, "logs the resume" on clear) and makes the PR's own manual-verification step (b) "validator logs the drain-mode entry" impossible to satisfy. Every sibling loop branch (empty/error/banned/transient) logs its state; this one should too (log-once on transition, not every poll).

**3. Retry queue is never flushed during drain.** `retry_queue.process_pending()` has one call site (main.py:571), reachable only after a successful claim+eval. The drain `continue` skips it, so previously-failed completion/score reports sit unsent for the entire (indefinite) drain window; if the instance terminates first, those results are lost and runs go TIMED_OUT/STALE on the backend.

**4. docker-compose bind mount contract.** `/var/run/oro-validator:/var/run/oro-validator` has no host-side pre-creation/ownership contract: Docker auto-creates it `root:root` (a non-root orchestrator then cannot `touch` the sentinel, so drain silently never engages), and `/var/run` is tmpfs (wiped on reboot, silently un-draining a node). Worth documenting/handling host-side.

**Lower severity (non-blocking):**
- Detection latency is TTL + poll_interval, not the documented ~10s — a stale False cache can mask a freshly-created file for a full TTL while a new long eval starts.
- `ORO_DRAIN_FILE` is read at import time, bypassing the argparse `get_config` pattern, so it is undiscoverable via `--help`/startup config dump.
- No test exercises the main-loop wiring (short-circuit + counter + sleep + in-flight finishes) — the behavior the PR delivers is untested; a refactor dropping the `continue` would stay green.

that-guy-wade:
Thanks for the deep review. Addressed all 4 blockers + 2 nits in 1262a03 (oro) + 3ad04f0 (Infrastructure PR #133):

**Blockers:**
1. **Auto-update racing drain** → drain gate now sits at the TOP of the loop body, above `_check_for_updates`. Watchtower can't restart the container mid-drain.
2. **No log on entry/clear** → log-once on entry ("Drain sentinel present...") and log-once on clear ("Drain sentinel cleared...") reinstated. Counter `CLAIM_WORK_TOTAL{result=draining}` still increments every tick.
3. **Retry queue never flushed** → `retry_queue.process_pending()` now runs inside the drain branch before the sleep, so pending completion/score reports leave the node before it terminates.
4. **Bind-mount host contract** → Infrastructure PR #133 adds a systemd-tmpfiles drop-in (`configs/oro-validator.tmpfiles.conf`) installed by bootstrap.sh, so `/var/run/oro-validator` exists at boot with explicit `root:root 0755` before docker compose brings the validator up. Handles both the /var/run tmpfs wipe AND the Docker auto-create root:root foot-gun.

**Nits:**
- Latency upper bound (cache TTL + poll_interval + claim_work response time) now documented in `drain.py` module docstring.
- New `test_main_loop_skips_claim_and_flushes_retry_queue_under_drain` directly exercises the loop wiring — claim_work and check_for_updates do NOT fire, retry_queue.process_pending DOES, counter increments. A future refactor that drops the `continue` or reorders the drain gate fails this test.
- `ORO_DRAIN_FILE` argparse exposure left as-is — env var + module-level constant is the same pattern the existing `AUTO_UPDATE_ENABLED` / `WATCHTOWER_URL` / `WATCHTOWER_TOKEN` constants follow. Happy to lift it into `get_config` in a follow-up if you'd prefer consistency in one direction.

shardi-b (changes requested):

## Blocking

**1. `test_main_loop_skips_claim_and_flushes_retry_queue_under_drain` is a tautology test (test_drain_mode.py:60).**
The test never imports or calls `Validator.run()` (or any code in `main.py`). It hand-rolls the loop body inline with local `MagicMock`s:

```python
if gate(cache, drain_file=drain_file):
    counter_inc(label="draining")
    if retry_queue.get_pending_count() > 0:
        retry_queue.process_pending()
else:
    backend.check_for_updates()
    backend.claim_work()
```

The only thing asserted is that this inline copy is internally consistent. Deleting the entire drain block from `main.py`, dropping the `continue`, or moving the auto-update gate above the drain check would all leave this test **green** — yet the docstring explicitly promises "This locks the wiring: a future refactor that drops the `continue` or moves the auto-update gate above the drain check will fail here." That promise is false. The drain-above-auto-update ordering is the entire safety rationale of the PR and currently has zero real test coverage, while the PR description cites the 5 passing tests as evidence the wiring works.

Please extract the loop iteration into a testable seam (e.g. a `_run_one_iteration`/`_should_drain` method on `Validator`) and assert against the real code path, or drive `run()` with mocked collaborators and a one-shot exit.

## Should fix

**2. Retry-queue reports can be dropped during drain (main.py:532).**
The drain branch calls `self.retry_queue.process_pending()` every `poll_interval` (default 30s). `process_pending` increments `retry_count` per transient `BackendError` and drops the entry after `max_retries=10` (retry_queue.py:92-98). During a multi-minute drain, a coincident backend transient (~5 min) exhausts all 10 retries and **permanently drops the pending completion/score reports** — the exact data this feature exists to flush before termination. Note the normal loop only calls `process_pending` after a *successful* eval cycle, so an outage there never burns retries; drain uniquely amplifies the drop. Consider not counting drain-time flush attempts against `max_retries`, or backing off the flush cadence.

**3. `drain_mode_active` fail-opens on unreadable sentinel path (drain.py:33).**
`os.path.exists` returns `False` (never raises) on EACCES/ENOTDIR. The new bind mount `- /var/run/oro-validator:/var/run/oro-validator` lets Docker auto-create the host dir root-owned, and `/var/run` is commonly tmpfs cleared on reboot. If the path is wrong-owned, replaced by a file, or the mount goes missing, the validator silently keeps claiming work despite a present sentinel, with no log/metric distinguishing "sentinel absent" from "cannot read sentinel path." Consider `os.stat` with explicit error logging so a misconfigured control path is observable.

## Minor

**4. Metric semantics (main.py:523).** `CLAIM_WORK_TOTAL{result="draining"}` increments every poll although `claim_work` is never called, contradicting the documented meaning "Outcome of claim_work polls" and skewing any success-rate dashboard.

**5. Dead test scaffolding (test_drain_mode.py:84).** The `with patch("subnet.validator.drain.DRAIN_FILE", ...)` block and re-import of `drain_mode_active as gate` are inert because `gate` is always called with an explicit `drain_file=` kwarg, so the patched module constant / env-var default is never exercised.

that-guy-wade:
Cycle 3 addressed in 171086f:

**1. Tautology test → real seam.** Extracted the loop body into `handle_drain_tick` as a free function in `drain.py` (also dodges main.py's bittensor import chain in tests). Tests now drive the actual function with mocked collaborators + a one-shot sleep_fn, asserting return value, `process_pending(count_attempts=False)` flush, metric increment, log-once on entry, log-once on resume.

**2. process_pending burning retries during drain.** Added `count_attempts: bool = True` kwarg. Normal post-eval flush keeps default (budget-counting); drain flush passes `count_attempts=False` so a multi-minute drain plus a coincident backend transient can't permanently drop completion/score reports before instance termination. The next successful post-drain iteration re-flushes against the unchanged budget.

**3. drain_mode_active fail-opens on unreadable path.** Switched from `os.path.exists` to `os.stat` with explicit `OSError` handling. EACCES / ENOTDIR / missing-mount → fail-CLOSED (return True) + WARNING. Covered by `test_drain_mode_active_fails_closed_on_eacces` (creates NotADirectoryError via parent-is-file).

**4. CLAIM_WORK_TOTAL{result=draining} skewing dashboards.** Reverted; added dedicated `DRAIN_TICKS_TOTAL` counter alongside the existing `CLAIM_WORK_TOTAL`. Claim-success rate stays clean; drain windows are observable on their own counter.

**5. Dead patch+re-import.** Removed (the whole tautology test went away).

All 10 drain tests + 179 validator tests pass locally.

shardi-b (dismissed):

shardi-b (approved):
Approved. No blocking issues found during code review.

The safety-critical change (retry_queue `process_pending` refactor) was traced carefully and is **behavior-preserving** for the default `count_attempts=True` path — identical to the original drop/append logic, no off-by-one. The `count_attempts=False` drain path correctly keeps reports without burning retry budget. Drain hook placement (above auto-update), the `state["logged"]` on/off transitions, and the lack of per-loop starvation (heartbeat/weight-setter run on independent threads) all check out.

A few non-blocking suggestions for a follow-up (none worth holding the merge):

1. `drain.py` `drain_mode_active`: consider `os.path.isfile()` instead of `os.stat()` so a directory created at the sentinel path is not treated as an active drain.
2. `drain.py` lazy `from .metrics import`: the comment says it prevents prometheus dual-registration, but `main.py` already eagerly imports both `.drain` and `.metrics`, so the lazy import is cosmetic — consider correcting the comment to avoid misleading a future maintainer.
3. `docker-compose.yml`: the `/var/run/oro-validator` mount could be `:ro` (matching the other config mounts) since the validator only reads the sentinel; the host writes it.
4. Fail-CLOSED on any `OSError` is the right default, but note a sustained transient FS error keeps the validator idle until cleared — worth an operator note.

Nice clean, well-tested, well-documented change.

PR #191, opened 2026-06-23
feat(validator): cap sandbox stdout/stderr capture (ORO-1414)

shardi-b:
## Description
Caps the validator's capture of sandbox stdout/stderr so a single miner agent can no longer fill the validator host's disk. This is the root-cause fix for the 2026-06-23 fleet incident where agent `kevgol` dumped ~14 GiB of stdout to one `sandbox_stdout.log`, filled the 48G root disk, and crashed the validator process (PD #257/#258). Because all validators pull from the same work queue, the same agent racked up **50 STALE runs across 7 validators** over ~6 hours. Discarding the agent stopped the bleed; this PR removes the capability.

## Changes Made
- Add `subnet/validator/bounded_io.py`:
  - `drain_capped(src, dst, max_bytes)` — copies a stream to disk writing at most `max_bytes`, drains the remainder so the child never blocks on a full pipe, and appends a truncation marker noting bytes dropped.
  - `run_capped(cmd, *, stdout_path, stderr_path, max_bytes, timeout)` — drop-in for `subprocess.run(stdout=f, stderr=f, timeout=...)`: `Popen` + two reader threads drain stdout/stderr concurrently through the cap. Preserves the `returncode` and re-raises `subprocess.TimeoutExpired` (killing the child) like `subprocess.run`.
- Wire `run_capped` into `_run_sandbox_inner` in `subnet/validator/main.py`, replacing the unbounded `subprocess.run(stdout=file, stderr=file)` capture.
- Add `--sandbox-log-max-bytes` arg (env `SANDBOX_LOG_MAX_BYTES`, default **256 MiB**).
- Side benefit: the later `stdout_log.read_text()` / `stderr_log.read_text()` calls can no longer OOM on a multi-GiB log, since the file is now bounded.

## Issue Link
- Related to: ORO-1414 (Follow-up 1: cap / rotate sandbox stdout capture)

### Manual Testing
N/A (covered by automated tests below; the wiring is a behavior-preserving swap verified by `py_compile` + ruff).

**Test Results:** 4 passed.

### Automated Testing
**Test Command(s):**
```bash
pytest subnet/tests/test_bounded_io.py -v
```
- `test_under_cap_writes_everything_verbatim` — output below cap is written verbatim, no marker.
- `test_over_cap_truncates_content_and_appends_marker` — only the cap's worth of content reaches disk; source still fully drained; truncation marker appended.
- `test_run_capped_bounds_a_runaway_stdout_to_disk` — a real subprocess spewing ~5 MB is bounded to ~4 KB on disk with the marker, returncode preserved.
- `test_run_capped_kills_on_timeout` — `subprocess.TimeoutExpired` raised and the child is killed at the timeout (not waited out).

Note: the full validator suite couldn't run in my local env (its conftest imports the heavier validator deps); CI runs it.

## Documentation
- [ ] README updated
- [x] Configuration documentation updated (new `--sandbox-log-max-bytes` / `SANDBOX_LOG_MAX_BYTES`, self-documented in `--help`)
- [ ] Other documentation updated

## Checklist
- [x] My changes generate no new warnings or errors (ruff clean, compiles)
- [ ] New and existing unit tests pass locally with my changes (new tests pass; full suite deferred to CI)

## Additional Notes
- Scope is intentionally narrow: this is the stdout/host-disk path only. The complementary docker-layer caps (`--storage-opt`, `--log-opt`) and the validator auto-restart and search-server backpressure are the other ORO-1414 follow-ups, tracked separately.
- Design choice: on cap exceedance the stream is still fully drained (rather than closing the pipe early) so the agent's process never blocks/SIGPIPEs mid-eval; the run completes normally with truncated logs.

### DIFF
--- a/subnet/validator/bounded_io.py
+++ b/subnet/validator/bounded_io.py
@@ -1,0 +1,103 @@
+"""Bounded capture of a sandbox output stream to disk (ORO-1414, Follow-up 1).
+
+A misbehaving agent can emit unbounded stdout/stderr. The validator captures
+that output to a host file, so the sandbox's own cgroup limits do not apply and
+a single run can fill the host disk (the kevgol incident: ~14 GiB to one
+``sandbox_stdout.log``). ``drain_capped`` keeps at most ``max_bytes`` on disk
+while still reading the source to EOF, so the child never blocks on a full pipe.
+"""
+
+from __future__ import annotations
+
+import subprocess
+import threading
+from pathlib import Path
+from typing import BinaryIO
+
+_DEFAULT_CHUNK_SIZE = 64 * 1024
+
+
+def drain_capped(
+    src: BinaryIO,
+    dst: BinaryIO,
+    max_bytes: int,
+    *,
+    chunk_size: int = _DEFAULT_CHUNK_SIZE,
+) -> int:
+    """Copy ``src`` to ``dst``, writing at most ``max_bytes`` of content.
+
+    Reads ``src`` to EOF regardless of the cap (draining) so the producer is
+    never blocked by a full pipe. If the source exceeds ``max_bytes``, a
+    truncation marker noting how many bytes were dropped is appended to ``dst``.
+
+    Returns the total number of bytes read from ``src``.
+    """
+    total = 0
+    written = 0
+    while True:
+        chunk = src.read(chunk_size)
+        if not chunk:
+            break
+        total += len(chunk)
+        if written < max_bytes:
+            to_write = min(len(chunk), max_bytes - written)
+            dst.write(chunk[:to_write])
+            written += to_write
+
+    if total > max_bytes:
+        dropped = total - max_bytes
+        dst.write(
+            f"\n...[output truncated: capped at {max_bytes} bytes, "
+            f"dropped {dropped} bytes]\n".encode()
+        )
+    return total
+
+
+def run_capped(
+    cmd: list[str],
+    *,
+    stdout_path: Path,
+    stderr_path: Path,
+    max_bytes: int,
+    timeout: float,
+    chunk_size: int = _DEFAULT_CHUNK_SIZE,
+) -> int:
+    """Run ``cmd``, capturing stdout/stderr to files each capped at ``max_bytes``.
+
+    Drop-in for ``subprocess.run(cmd, stdout=f, stderr=f, timeout=...)`` that
+    bounds how much reaches disk. Reader threads drain both pipes concurrently
+    (so the child never blocks on a full pipe and stdout/stderr can't deadlock).
+    Raises ``subprocess.TimeoutExpired`` and kills the child on timeout, matching
+    ``subprocess.run``. Returns the child's exit code.
+    """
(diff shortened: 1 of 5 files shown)


subnet/validator/main.py
```diff
@@ -343,24 +354,24 @@ def _run_sandbox_inner(
...
+            returncode = run_capped(
+                cmd,
+                stdout_path=stdout_log,
+                stderr_path=stderr_log,
+                max_bytes=self.config.sandbox_log_max_bytes,
+                timeout=self.config.sandbox_timeout,
+            )
+            metadata["exit_code"] = returncode
 
             # Always log sandbox output for debugging
```
that-guy-wade, line 370: Pre-existing risk (not introduced here): if an agent emits non-UTF-8 bytes to stdout/stderr, `read_text()` raises `UnicodeDecodeError` and we lose `stderr_tail` + the debug log. Worth a follow-up to use `read_text(errors='replace')` (or `read_bytes().decode(errors='replace')`) so the bound is useful even when the agent writes binary.

subnet/validator/bounded_io.py
```diff
@@ -0,0 +1,103 @@
...
+            )
+            for pipe, dst in ((proc.stdout, out_file), (proc.stderr, err_file))
+        ]
+        for t in readers:
+            t.start()
+        try:
+            proc.wait(timeout=timeout)
+        except subprocess.TimeoutExpired:
+            proc.kill()
+            proc.wait()
```
that-guy-wade: Heads-up (not a regression): `proc.kill()` SIGKILLs the immediate child, which in our path is the `docker` CLI. The sandbox container itself keeps running on the daemon until its own timeout. Output written after the CLI dies goes to /dev/null so the on-disk cap still holds, but the container can keep burning CPU/network briefly. Same behavior as the pre-PR `subprocess.run`, just worth being aware of when reasoning about timeouts.

that-guy-wade (dismissed):
Approved — clean, well-scoped, tight tests. Two minor nits for follow-up (not blocking).

subnet/validator/main.py
shardi-b: Fixed in 024dfe7. Added `read_text_lossy` (`read_text(errors="replace")`) and routed all four sandbox-log read sites through it, so a binary-spewing agent no longer loses the stderr_tail or debug log. Test: `test_read_text_lossy_tolerates_non_utf8`.

subnet/validator/bounded_io.py
shardi-b: Addressed in 5324238. The sandbox container now gets a deterministic `--name` (`oro-sandbox-<eval_run_id>`) and we `docker kill` it on the timeout path so the orphan stops burning CPU/network instead of waiting on its own internal timeout.

Self-review of that change surfaced two leak gaps in the same disk-full path, fixed in 7f3c1d1: `run_capped` now reaps the child if the cap-file `open()` raises after `Popen` (ENOSPC on an already-full disk), and the generic `except Exception` in `_run_sandbox_inner` also kills the orphan now (previously only `TimeoutExpired` did).

that-guy-wade (approved):
LGTM

PR #192, opened 2026-06-23
feat(validator): in-process watchdog to recover a wedged validator (ORO-1414)

shardi-b:
## Description
Adds an in-process liveness watchdog so a wedged validator auto-recovers. The container already has `restart: unless-stopped`, but that only fires on a clean **exit**. In the 2026-06-23 incident the process wedged when the host disk filled (it stopped making progress without exiting), so the container stayed `Up` with `0B` in `docker stats` and never recovered until a manual `systemctl restart` ~90 min later (PD #257/#258). This is ORO-1414 Follow-up 2.

## Changes Made
- Add `subnet/validator/watchdog.py` — `ProgressWatchdog`: a daemon thread that aborts the process via `os._exit(1)` when the main loop hasn't beaten within the timeout, so the restart policy brings a fresh validator up. `os._exit` (not `sys.exit`) bypasses cleanup/atexit so a wedged process can't swallow the exit.
- Wire it into `run()`: the main loop calls `watchdog.beat()` at the top of each iteration; started after the weight setter, stopped in the shutdown `finally`.
- Add `--validator-watchdog-timeout` (env `VALIDATOR_WATCHDOG_TIMEOUT`, default **5400s = 90 min**; `0` disables).

## Design notes
- **No false-kills:** the beat is once per outer-loop iteration, so the timeout must exceed the longest legitimate iteration. Worst case is a hung update check (~900s) + a full eval (sandbox 1800s + scoring/reasoning ~900s+) ≈ 3600–3900s; the 5400s default leaves margin. A true total wedge is recycled within ~90 min.
- **Tighter recovery (follow-up):** beating from the heartbeat thread (which already fires every 30s *during* an eval) would decouple the timeout from eval duration and allow a much shorter timeout. Deferred because `HeartbeatManager` pulls in deps I couldn't verify locally; tracked on ORO-1414.
- **Accepted limitation:** a hard interpreter hang (GIL deadlock) starves the watchdog thread too; this catches the realistic "loop stopped progressing" case (the incident), not a full deadlock.
- **`os._exit` skips the in-flight eval-dir cleanup.** Bounded: PR #191 caps each log at 256 MiB and the watchdog fires only on a real wedge, so a leaked dir is small. A startup sweep of orphaned `eval_*` dirs is a sensible related follow-up (also closes an ORO-675 gap).

## Issue Link
- Related to: ORO-1414 (Follow-up 2: auto-restart the validator on a wedge)

### Automated Testing
**Test Command(s):**
```bash
pytest subnet/tests/test_watchdog.py -v
```
5 passed: not-stalled-before-timeout, stalled-after-timeout, beat-resets-the-timer, monitor-thread-aborts-when-stalled (injected `on_stall`), monitor-thread-quiet-while-healthy. Pure unit via an injectable clock — no real waiting, deterministic.

Note: the full validator suite couldn't run in my local env (its conftest needs the heavier validator deps); CI runs it. `py_compile` + ruff clean on the changed files.

## Documentation
- [ ] README updated
- [x] Configuration documentation updated (new `--validator-watchdog-timeout` / `VALIDATOR_WATCHDOG_TIMEOUT`, self-documented in `--help`)

## Checklist
- [x] My changes generate no new warnings or errors (ruff clean, compiles)
- [x] I have added tests that prove my fix works
- [ ] New and existing unit tests pass locally (new tests pass; full suite deferred to CI)

## Additional Notes
- Independent of #191 (the stdout cap) — both are ORO-1414 follow-ups but touch different paths; either can merge first.

### DIFF
--- a/subnet/validator/watchdog.py
+++ b/subnet/validator/watchdog.py
@@ -1,0 +1,90 @@
+"""In-process liveness watchdog for the validator main loop (ORO-1414, Follow-up 2).
+
+The validator container has ``restart: unless-stopped``, which only fires when
+the container *exits*. In the 2026-06-23 incident the process wedged (host disk
+filled) without exiting, so the container stayed ``Up`` (0B) and never
+restarted. This watchdog turns a stalled main loop into a process exit: the
+loop calls :meth:`beat` each iteration, and a background thread aborts the
+process if no beat lands within ``timeout_seconds`` so the restart policy can
+bring a fresh validator back up.
+
+Known limitation (accepted): if the whole interpreter is hard-hung (e.g. GIL
+deadlock), the watchdog thread itself is starved and cannot fire.
+"""
+
+from __future__ import annotations
+
+import logging
+import os
+import threading
+import time
+from typing import Callable
+
+# The loop beats once per iteration, so the default must exceed the longest
+# legitimate single iteration. Worst case is a hung update check (~900s) plus a
+# full eval (sandbox_timeout 1800s + scoring/reasoning ~900s+) in the same pass,
+# ~3600-3900s; 5400s leaves margin so a healthy busy validator never trips it,
+# while a truly wedged loop is still recycled within ~90 min. Tighter recovery
+# would require beating from the heartbeat thread (fires every 30s during an
+# eval) — see ORO-1414 follow-up.
+_DEFAULT_TIMEOUT_SECONDS = 5400.0
+_DEFAULT_CHECK_INTERVAL = 30.0
+
+
+class ProgressWatchdog:
+    """Abort the process if the main loop stops beating within ``timeout_seconds``."""
+
+    def __init__(
+        self,
+        timeout_seconds: float = _DEFAULT_TIMEOUT_SECONDS,
+        *,
+        check_interval: float = _DEFAULT_CHECK_INTERVAL,
+        now: Callable[[], float] = time.monotonic,
+        on_stall: Callable[[], None] | None = None,
+    ) -> None:
+        self._timeout = timeout_seconds
+        self._check_interval = check_interval
+        self._now = now
+        self._on_stall = on_stall or self._abort
+        self._last_beat = now()
+        self._stop = threading.Event()
+        self._thread: threading.Thread | None = None
+
+    def beat(self) -> None:
+        """Record forward progress; resets the stall timer."""
+        self._last_beat = self._now()
+
+    def stalled(self) -> bool:
+        return (self._now() - self._last_beat) > self._timeout
+
+    def start(self) -> None:
+        if self._thread is not None:
+            return
+        self._thread = threading.Thread(
+            target=self._run, name="validator-watchdog", daemon=True
+        )
+        self._thread.start()
+
+    def stop(self) -> None:
+        self._stop.set()
+        if self._thread is not None:
+            self._thread.join(timeout=self._check_interval + 1.0)
+            self._thread = None
+
+    def _run(self) -> None:
+        while not self._stop.wait(self._check_interval):
+            if self.stalled():
+                self._on_stall()
+                return
+
+    def _abort(self) -> None:
+        elapsed = self._now() - self._last_beat
+        logging.error(
+            "Validator watchdog: no progress for %.0fs (timeout %.0fs) — aborting "
+            "so the container restart policy recovers the validator (ORO-1414).",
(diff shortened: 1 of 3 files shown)


subnet/validator/watchdog.py
```diff
@@ -0,0 +1,90 @@
...
+        elapsed = self._now() - self._last_beat
+        logging.error(
+            "Validator watchdog: no progress for %.0fs (timeout %.0fs) — aborting "
+            "so the container restart policy recovers the validator (ORO-1414).",
+            elapsed,
+            self._timeout,
+        )
+        # os._exit, not sys.exit: bypass cleanup/atexit so a wedged process can't
+        # swallow the exit; the container exits and `restart: unless-stopped` fires.
+        os._exit(1)
```
that-guy-wade, line 95: Not blocking — worth a follow-up: the watchdog only sees the *main loop*. If `WeightSetter` (or another long-running background thread) wedges while the main loop keeps spinning, the beat keeps landing and we never recover. The PR design note already calls out beating from the heartbeat thread as the tighter-recovery move — same fix would let us also detect background-thread wedges if each contributing thread has to beat.

subnet/validator/watchdog.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def _abort(self) -> None:
+        elapsed = self._now() - self._last_beat
+        logging.error(
+            "Validator watchdog: no progress for %.0fs (timeout %.0fs) — aborting "
+            "so the container restart policy recovers the validator (ORO-1414).",
+            elapsed,
+            self._timeout,
+        )
+        # os._exit, not sys.exit: bypass cleanup/atexit so a wedged process can't
```
that-guy-wade, line 93: Optional follow-up: bump a Prometheus counter (or CloudWatch metric) right before `os._exit` so we can detect validator restart loops in monitoring. Today a self-recovery is invisible to alerting unless someone is reading container logs. Could piggyback on the existing metrics module.

that-guy-wade (dismissed):
Approved — clean follow-up to #191, mirrors the same quality bar. Two follow-up nits inline (not blocking).

subnet/validator/watchdog.py
shardi-b: Good catch. Leaving this as a tracked follow-up rather than rushing it in: a correct fix needs each long-running thread (WeightSetter, heartbeat) to beat its own source, and the consumer wiring is the part I can't verify locally (the validator conftest needs the heavier deps, so a false-kill in a safety mechanism would go uncaught). Tracked on ORO-1414 alongside the heartbeat-beat tighter-recovery item. This PR keeps the main-loop-only scope it was approved with.

subnet/validator/watchdog.py
shardi-b: Addressed in d458675, but not as a Prometheus counter. The process `os._exit`s immediately inside `_abort`, before the next scrape, and the counter resets to 0 on restart, so a scrape-based counter never surfaces the event. Instead the abort log now carries a stable `WATCHDOG_ABORT` token so a CloudWatch Logs metric filter + alarm can catch validator restart loops. Test stubs `os._exit` and asserts the token is logged.

that-guy-wade (dismissed):
LGTM

that-guy-wade (approved):
Approving. Right shape for the wedge problem:
- `os._exit(1)` (not `sys.exit`) is the correct choice — a wedged process can't swallow it.
- Beat-at-top-of-outer-loop avoids false-kills from slow legit evals.
- 5400s default math (update ~900s + sandbox 1800s + scoring ~900s) checks out.
- `WATCHDOG_ABORT` stable token is the right observability handle for a CW metric filter alarm.
- Injectable `now` + `on_stall` gives clean deterministic tests.

One nit (non-blocking — happy to land as-is if it's already imported):

**`Optional` import in main.py.** The new line `watchdog: Optional[ProgressWatchdog] = None` is at `subnet/validator/main.py:~570`. The diff doesn't add `from typing import Optional`, so this is a `NameError` at startup if it's not already imported in main.py. Two cheap options:
- Confirm `Optional` is already imported in main.py (it probably is — the file is large).
- Or swap to `watchdog: ProgressWatchdog | None = None` (works on the 3.10+ runtime the validator already targets).

Acknowledged limitations in the PR description (hard interpreter hang, eval-dir cleanup on abort) are fine to ship — both are realistic-case-first decisions.

PR #195, opened 2026-06-24
sandbox: raise --cpu-shares 512→2048 and ulimit nofile 1024→4096

that-guy-wade:
## Description

Two related sandbox resource throttles surfaced while testing `SANDBOX_MAX_WORKERS=60` on a small validator host (c7g.2xlarge, 8 vCPU). Both kicked in only under saturation; on idle hosts behavior is unchanged.

## Changes Made

- **`subnet/sandbox.py`**: bump `--cpu-shares` 512 → 2048 (sandbox no longer starved during its own phase) and `nofile` ulimit 1024 → 4096 (no silent FD-cap on high-concurrency agents).
- **`subnet/tests/test_sandbox.py`**: updated the two assertions that pin the old values.

## Issue Link

- Related to: N/A
- Closes: N/A

### Manual Testing

Replayed the same prod race winner (race 70's `octopus`) on a staging validator at `SANDBOX_MAX_WORKERS=60` before and after the change, on a c7g.2xlarge host with 8 vCPU. Observed:

- **Before:** sandbox averaged 131 s per problem; 27 of 120 problems hit the 300 s per-problem timeout.
- **After:** TBD — will measure on the same host with the new image and update this PR.

### Automated Testing

- `subnet/tests/test_sandbox.py::test_resource_limits` updated for the new values and passes locally.

**Test Command(s):**
```bash
uv run pytest subnet/tests/test_sandbox.py -q
```

## Documentation

- [ ] README updated — N/A
- [x] Code comments added/updated — yes, both blocks now explain the rationale (subprocess.run() blocks the validator main loop during sandbox, so giving sandbox more CPU share is safe; nofile 1024 was a silent cap at 60-worker fan-out).
- [ ] API documentation updated — N/A
- [ ] Configuration documentation updated — N/A

**Documentation Changes:** Inline comments only.

## Checklist

## Additional Notes

Cost analysis:
- `cpu-shares` is relative weight only — under contention sandbox gets ~2/3 of CPU instead of ~1/7; with no contention, behavior is identical.
- `nofile` increase costs ~1 KB kernel memory per FD = ~3 MB peak under load. `--pids-limit 256` and `--memory 4g` still catch runaway processes.

Both flags only matter when the host is saturated; idle hosts unaffected.

### DIFF
--- a/subnet/sandbox.py
+++ b/subnet/sandbox.py
@@ -165,12 +165,17 @@ def build_sandbox_command(
         "--pids-limit",
         "256",
         "--ulimit",
-        "nofile=1024:1024",
-        # CPU priority — validator + search-server (default cpu-shares=1024) preempt the
-        # sandbox under contention so heartbeats and the work-claim loop stay responsive.
-        # Sandbox still uses idle CPU; this only matters when the host is saturated.
+        # 4096 fds: at SANDBOX_MAX_WORKERS=60 a typical agent holds ~5 sockets
+        # in flight per worker (HTTP + DNS + keepalive), pushing toward the
+        # 1024 cap and silently capping throughput before the worker count.
+        "nofile=4096:4096",
+        # CPU priority — sandbox runs during the validator's blocked
+        # subprocess.run(); only the heartbeat + weight-setter threads compete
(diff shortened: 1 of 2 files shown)


shardi-b (approved):

PR #196, opened 2026-06-25
validator: parallelize per-problem log upload + reuse TCP connection

that-guy-wade:
## Description

The post-sandbox phase uploads one gzipped JSONL per problem to S3. At N=120 problems that's a sequential loop of 120 × (backend presign + S3 PUT), and each call opens a fresh TCP+TLS connection because the code calls ``requests.request`` at module scope rather than a Session. Measured on a live 120-problem octopus eval, this single loop burns **~75 seconds** of wall time (628 ms / problem) — about 15% of the whole eval.

## Changes Made

- `subnet/validator/backend_client.py`: introduce a module-level `_S3_SESSION = requests.Session()` and route `upload_to_s3` through it so urllib3 keeps connections alive between problems. Default pool size (10) covers the parallel uploader below.
- `subnet/validator/main.py`: `_upload_logs` now fans the per-problem (presign + PUT) work out over a 20-thread `ThreadPoolExecutor` instead of a serial loop. Each worker calls `backend_client.get_presigned_upload_url` (httpx; SDK has its own pool) and then `upload_to_s3` (requests Session above). At 20 workers + 10 S3 sockets, urllib3 lightly queues but every PUT still reuses TCP.
- `subnet/tests/validator/test_backend_client.py`: update patch target from `requests.request` to `_S3_SESSION.request` to match the new call site. 32 backend_client tests pass.

## Issue Link

- Related to: N/A — depends on `Backend#433` for matching rate-limit headroom
- Closes: N/A

### Manual Testing

Profiled a live 120-problem eval on staging to measure where the post-sandbox wall time goes; identified the sequential upload loop as the dominant non-sandbox bucket. Estimated speedup: ~75 s → ~5 s based on observed per-call latency × 20 parallelism.

### Automated Testing

`subnet/tests/validator/test_backend_client.py` — 32 tests pass, including the two upload-to-S3 unit tests that now target the new Session.

**Test Command(s):**
```bash
uv run pytest subnet/tests/validator/test_backend_client.py -q
```

## Documentation

- [ ] README updated — N/A.
- [x] Code comments added/updated — yes, both the module-level `_S3_SESSION` and the new `_UPLOAD_LOGS_WORKERS` constant carry rationale comments.
- [ ] API documentation updated — N/A.
- [ ] Configuration documentation updated — N/A.

**Documentation Changes:** inline only.

## Checklist

## Additional Notes

**Backend rate-limit dependency.** A burst of 20 concurrent presign requests from one IP would have tripped the old `100/min` global IP cap. `Backend#433` raises the per-IP cap on `/v1/validator/*` paths to `600/min` while keeping the bucket keyed on the unforgeable client IP. This PR is safe to ship once that Backend change is deployed to staging — sequencing matches the PR description there.

**Thread-safety.**
- `requests.Session` is thread-safe for concurrent requests against the same Session instance (urllib3 PoolManager handles concurrency).
- `backend_client.get_presigned_upload_url` flows through the SDK's httpx client which is async-batched internally and tolerates parallel invocations from threads via the SDK's sync wrappers.
- The closure `_upload_one` only reads `eval_run_id` (an immutable UUID) and the BackendClient (used for its thread-safe methods).

**Sizing.** 20 workers chosen as a round number that matches typical urllib3 pool defaults and leaves a 2× margin under the Backend's raised 600/min IP cap (~270 RPM observed sustained during a 120-problem eval).

### DIFF
--- a/subnet/tests/validator/test_backend_client.py
+++ b/subnet/tests/validator/test_backend_client.py
@@ -312,7 +312,7 @@ class TestBackendClientUploadToS3:
         mock_response.status_code = 200
 
         with patch(
-            "validator.backend_client.requests.request", return_value=mock_response
+            "validator.backend_client._S3_SESSION.request", return_value=mock_response
         ):
             client.upload_to_s3(presign, b"test data")  # Should not raise
 
@@ -331,7 +331,7 @@ class TestBackendClientUploadToS3:
         mock_response.status_code = 403
 
         with patch(
-            "validator.backend_client.requests.request", return_value=mock_response
+            "validator.backend_client._S3_SESSION.request", return_value=mock_response
         ):
             with pytest.raises(BackendError):
                 client.upload_to_s3(presign, b"test data")
--- a/subnet/validator/backend_client.py
+++ b/subnet/validator/backend_client.py
@@ -175,6 +175,13 @@ class BackendError(Exception):
         return self.message
 
 
+# Long-lived Session used by ``upload_to_s3`` so the validator's per-eval
+# upload burst reuses TCP+TLS connections instead of opening a fresh socket
+# per problem. Default urllib3 pool size (10) is enough to keep the
+# parallel uploader in _upload_logs cache-warm.
+_S3_SESSION = requests.Session()
+
+
 class BackendClient:
     """HTTP client for ORO Backend API using oro-sdk.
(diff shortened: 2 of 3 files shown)


shardi-b (approved):

PR #197, opened 2026-06-25
validator: double default workers (sandbox 15→30, reasoning 4→8)

that-guy-wade:
## Description

Bump validator default worker counts to double current throughput:
- `SANDBOX_MAX_WORKERS` default: 15 → 30
- `REASONING_MAX_WORKERS` default: 4 → 8

Env-var overrides still take precedence — hosts that explicitly set either value see no change.

## Changes Made

- `subnet/validator/main.py`: bumped the two argparse defaults for `--sandbox-max-workers` and `--reasoning-max-workers`.

## Issue Link

- Related to: N/A
- Closes: N/A

### Manual Testing

Staging has been running real production agents at 120 sandbox workers (8× the old default, 4× this PR's new default) with no issues observed — 30 leaves plenty of headroom on the `c7g.2xlarge`.

Pre-merge: confirmed neither `SANDBOX_MAX_WORKERS` nor `REASONING_MAX_WORKERS` is set in the prod validator host `.env`, so new defaults will take effect on next image roll. Staging `.env` has both set (60 / 16) and is unaffected.

**Test Results:**

Staging confirmed stable at 4× new default. Prod will move from 15/4 → 30/8 on next image roll.

### Automated Testing

No defaults-asserting tests exist, and `subnet/sandbox.py` fd-limit code already sized for up to 60 workers, so no new tests needed.

**Test Command(s):**

```bash
uv run pytest subnet/tests
```

## Documentation

- [ ] README updated

**Documentation Changes:**

N/A — argparse `help=` text already documents env-var overrides and is unchanged.

## Checklist

## Additional Notes

Validator host = `c7g.2xlarge` (8 vCPU, 16 GB). Staging running 4× the new default with real agents demonstrates the headroom.

## Sandbox memory headroom (testing data, 2026-06-24)

Captured during a series of staging evals at `SANDBOX_MAX_WORKERS=60` (4× this PR's new default). Container is launched with `--memory 4g --memory-swap 4g`.

```
docker stats — Mac sandbox container readings (--memory 4g cap):
  765 MiB / 4 GiB  (19%)  — startup, before scoring kicked in
  1.04 GiB / 4 GiB (26%)  — early problem execution
  1.126 GiB / 4 GiB (28%) — mid-eval
  1.156 GiB / 4 GiB (29%) — mid-eval (different point)
  1.199 GiB / 4 GiB (30%) — observed peak across snapshots
```

**Peak observed: ~1.2 GiB out of 4 GiB cap (~30% utilization, ~2.8 GiB headroom).**

Caveat: `docker stats` samples instantaneously, so the true peak between samples may be slightly higher — best estimate is 1.3–1.6 GiB based on the consistency of the snapshots. Still well below the 4 GiB cap.

For the new 30-worker default this PR ships, per-eval peak should be roughly half of the 60-worker observations (~600–800 MiB), leaving ~3+ GiB of headroom on the existing cap.

### DIFF
--- a/subnet/validator/main.py
+++ b/subnet/validator/main.py
@@ -139,7 +139,7 @@ class Validator:
         parser.add_argument(
             "--sandbox-max-workers",
             type=int,
-            default=int(os.environ.get("SANDBOX_MAX_WORKERS") or "15"),
+            default=int(os.environ.get("SANDBOX_MAX_WORKERS") or "30"),
             help="Number of parallel problem workers in sandbox (env: SANDBOX_MAX_WORKERS).",
         )
         parser.add_argument(
@@ -151,7 +151,7 @@ class Validator:
         parser.add_argument(
             "--reasoning-max-workers",
             type=int,
-            default=int(os.environ.get("REASONING_MAX_WORKERS") or "4"),
+            default=int(os.environ.get("REASONING_MAX_WORKERS") or "8"),
             help="Number of parallel reasoning judge workers (env: REASONING_MAX_WORKERS).",
         )
         # Backend API configuration


shardi-b (dismissed):

## Finding 1 (blocking): `--memory 4g` cap not scaled with the 2x worker bump

The 30 sandbox workers all run inside a **single container** hard-capped at `--memory 4g --memory-swap 4g` (`subnet/sandbox.py:161-164`). `memory-swap == memory` means swap is fully disabled. Doubling workers halves the per-agent memory budget: **~273 MB/worker at 15 -> ~137 MB/worker at 30**.

An OOM-killed agent subprocess is scored as a miner fault: in `src/agent/sandbox_executor.py:337-347` a SIGKILLed child (exitcode -9, empty result queue) returns `status=FAILED`, indistinguishable from a real crash. A memory-heavy agent that passed at 15 workers now gets cgroup-OOM-killed at 30 and is scored 0 -- corrupting the leaderboard/emissions. This is the exact OOM the PR body flags ("xlarge OOMed at lower concurrency historically").

The sibling `--ulimit nofile` and `--cpu-shares` were explicitly re-sized for 60 workers in recent commits, but `--memory` was left untouched. The mitigation ("watch RAM on staging after rollout") validates *after* merge -- for a leaderboard-affecting risk this should gate the merge.

**Suggested:** either bump `--memory` proportionally (and re-check `--tmpfs /tmp size=256m`, which also counts against the cgroup and is shared across all 30 workers), or land the staging RAM validation *before* merge rather than after.

## Secondary (non-blocking, worth noting)
- **Judge 429 amplification:** 8 vs 4 reasoning workers double concurrent judge calls on a single shared miner token; code already warns about rate contention (`reasoning_scorer.py:24-31`) and a 402 trips a circuit breaker disabling reasoning scoring for the whole eval (`reasoning_judge.py:73-79`).
- **Scoring backpressure:** 30-wide sandbox feeds an 8-wide scoring pool with an unbounded futures dict (`scoring_pool.py:54`); bursts can trip `mark_remaining_timed_out()` (900s) -> problems marked TIMED_OUT.

## Verified OK
- pids-limit 256: workers are threads + one shared forkserver helper + one child per problem (~32 tasks at 30 workers), so 256 has headroom. Not a concern.
- fd/cpu-shares plumbing genuinely sized for 60 workers. No stale doc/compose references; docker-compose uses empty-fallback env vars.

that-guy-wade:
Note: staging has been running real production agents at 120 sandbox workers (8× the old default, 4× this PR's new default) with no issues observed — so 30 leaves plenty of headroom on the c7g.2xlarge.

shardi-b (approved):

PR #202, opened 2026-06-26
docker: bundle py-spy (validator) + async-profiler (search-server)

that-guy-wade:
## Summary

Bundle two profilers in the images so we can flame-graph the hot path inside each container without restarting the service or rebuilding from scratch. Both are zero overhead at runtime — only active when explicitly invoked.

Needed because the public host-metrics snapshot tells us total container CPU but not which process or stack frame is burning it. External validators (e.g. tao.bot at sustained ~96% CPU) cost us scoring quality and we have no way to attribute it.

## Changes

- `docker/validator/pyproject.toml` + `uv.lock`: pin `py-spy>=0.4.1`. Lock regenerated via `uv lock`.
- `docker/search-server/Dockerfile`: download `async-profiler` 3.0 release tarball for x64/arm64, symlink `asprof` onto `PATH`. ~5 MiB added per arch.

## How to use

**Validator (Python):**
```bash
docker exec oro-validator-1 .venv/bin/py-spy dump --pid 1
docker exec oro-validator-1 .venv/bin/py-spy record -o /tmp/v.svg --pid 1 -d 30
docker cp oro-validator-1:/tmp/v.svg .
```

**Search-server (JVM):**
```bash
docker exec shoppingbench-search-server asprof -d 30 -f /tmp/s.html 1
docker cp shoppingbench-search-server:/tmp/s.html .
```

Open the resulting `.svg` / `.html` in a browser.

## Test plan

- [ ] CI green (lint, tests, image build for both validator + search-server).
- [ ] After tagged release + rollout to staging, verify `docker exec` invocations succeed against a running staging validator + search-server.
- [ ] No change to runtime behavior with the tools idle.

### DIFF
--- a/docker/search-server/Dockerfile
+++ b/docker/search-server/Dockerfile
@@ -2,6 +2,26 @@ FROM ghcr.io/oro-ai/oro/search-server-base:latest
 
 WORKDIR /app
 
+# Bundle async-profiler so we can JVMTI-attach to the embedded Lucene JVM
+# (pyserini) at runtime without restarting the container. ~5 MiB; opt-in
+# only — never invoked unless an operator runs `asprof` manually inside
+# the container.
+ARG ASPROF_VERSION=3.0
+RUN set -eux; \
+    arch="$(uname -m)"; case "$arch" in \
+        x86_64) asprof_arch=x64 ;; \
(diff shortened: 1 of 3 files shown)


shardi-b (approved):

PR #203, opened 2026-06-27
scoring: swap rule-score embedder to bge-small-en-v1.5

that-guy-wade:
## Description

Replace `Qwen/Qwen3-Embedding-0.6B` (596M params, bf16) with `BAAI/bge-small-en-v1.5` (33M params, fp32) as the embedder used inside `rule_score_reward` for product-title similarity. Adjusts the title-similarity threshold from `0.7` to `0.95` to match calibrated agreement between the two models.

Qwen3-Embedding-0.6B was the single largest CPU consumer in validator scoring (~74% of scoring CPU on flame-graph traces). bge-small reduces embedding wall-time by ~25× while preserving the strict `rule >= 1` problem-success gate.

## Changes Made

- `src/agent/rewards/orm.py`: extract `SENTENCE_MODEL_NAME` + `TITLE_SIM_THRESHOLD` constants; load `BAAI/bge-small-en-v1.5`; bump threshold `0.7 → 0.95`.
- `docker/validator/Dockerfile`: pre-cache `BAAI/bge-small-en-v1.5` instead of `Qwen/Qwen3-Embedding-0.6B`.
- `tests/test_scoring_perf.py`: update inline comment to reference `TITLE_SIM_THRESHOLD`.

## Issue Link

- Related to: N/A
- Closes: N/A

### Manual Testing

Reproducer evaluations on a testnet validator using three pre-existing agents, qualifying suite (60 problems):

| Agent | bge-small @ 0.95 | Qwen3 baseline |
|-------|------------------|----------------|
| A | 13/60 success (369s) | 15/60 success (462s) |
| B | 8/60 success (395s) | 10/60 success (492s) |
| C | 9/60 success (388s) | 11/60 success (499s) |

Per-problem flip analysis isolating same-chosen-product cases across 107 controlled problems found only ±1 binary outcome difference attributable to the embedder switch; the remaining variance is agent-output stochasticity (≈±4 successes per agent run-to-run).

**Test Results:** Average runtime reduction ≈21% per evaluation. Scoring outcome within run-to-run noise of the previous embedder.

### Automated Testing

Existing unit tests for `rule_score_reward` updated and pass.

**Test Command(s):**

```bash
.venv/bin/python -m pytest tests/ subnet/tests/
```

457 passed, 5 skipped.

## Documentation

- [ ] README updated

**Documentation Changes:** N/A

## Checklist

## Additional Notes

Threshold `0.95` was calibrated against a 3600-record cross-model agreement sweep on prior race data: at this value bge-small reproduces Qwen3 binary `rule >= 1` outcomes with ≈97.5% agreement.


shardi-b (dismissed):

**[High] Dimension/model mismatch with precomputed `reward_title_embeddings` (`src/agent/rewards/orm.py:78`)**

`rule_score_reward` mixes embeddings from two different models in a single `model.similarity` call:

- `product_emb` is always encoded by the **new** model (bge-small, 384-dim) via `model.encode(...)` / `batch_encode_titles(...)` (`problem_scorer.py:283`).
- On the non-GT title path, `gt_emb` can come from `reward["_title_embeddings"]` (`orm.py:73`), populated from the problem suite's `reward_title_embeddings` (`scoring_pool.py:93-94` → `attach_title_embeddings`, `subnet/sandbox.py:85`). Those dataset embeddings were precomputed with the **old** model, Qwen3-Embedding-0.6B (1024-dim).

When a recommended product is not a GT match but its reward title is in `precomputed`, the call becomes `model.similarity((1,384), (1,1024))`, which raises a matmul shape-mismatch `RuntimeError` and crashes scoring. Even if dims aligned, cross-model embedding comparison is semantically meaningless → garbage scores.

The testnet reproducer passing only indicates those suites lacked `reward_title_embeddings`; production suites that ship precomputed embeddings would break. The unchecked "dependent changes published and merged" box reinforces that dataset-side embedding regeneration has not happened.

**Fix:** regenerate and re-publish the problem-suite `reward_title_embeddings` with bge-small as a dependent change in lockstep with this swap, OR add a dimension guard in `rule_score_reward` that falls back to re-encoding when a precomputed embedding's dimension does not match the active model.

that-guy-wade:
Thanks for the catch — that was a real bug. Production problem suites do ship `reward_title_embeddings` precomputed with Qwen3 (1024-dim), and the unconditional `precomputed[title]` lookup would have either crashed on `model.similarity` shape-mismatch or silently fed cross-model embeddings into the cosine.

Pushed `e725fbf`: compare each cached embedding's `len` against `model.get_sentence_embedding_dimension()` and fall back to live `model.encode(...)` when they disagree. Suites that later ship matching-dim (bge-small) embeddings keep the fast path; current Qwen3-embedded suites transparently fall back.

Added a unit test (`test_stale_precomputed_embedding_dim_mismatch_falls_back`) that asserts: when a 1024-dim cached embedding is present and the live model dim is 384, we live-encode both sides and `similarity` only ever sees matching-dim tensors.

shardi-b (approved):

PR #194, opened 2026-06-24
validator memory: bf16 embedding model + JVM heap cap

that-guy-wade:
## Description

Two small, independent reductions to validator-host memory footprint. Both ship together for a combined ~3 GiB savings on validator memory.

## Changes Made

- **bf16 embedding model** (`src/agent/rewards/orm.py`): switch `Qwen3-Embedding-0.6B` from `float32` to `bfloat16`. Qwen3 family is bf16-native (training dtype), so this is precision-equivalent for the cosine-similarity threshold at 0.7. Saves ~1.2 GiB resident on the validator process.
- **JVM heap cap on search-server** (`docker/search-server/Dockerfile` + `docker-compose.yml`): pin pyserini's embedded JVM to `-Xmx2g`. The previous Dockerfile set `JAVA_OPTS` (non-standard, ignored by pyserini/jnius) — the JVM was running on defaults. Switching to `_JAVA_OPTIONS` (read directly by the JVM at startup) gives a real 2 GiB cap. Compose passes through `_JAVA_OPTIONS` so per-host overrides remain possible.

## Issue Link

- Related to: N/A
- Closes: N/A

### Manual Testing

**Memory savings — bf16 model:**
1. Built the validator image with the bf16 change locally.
2. Ran a validator container; confirmed the embedding model loads in ~1.2 GiB (was ~2.4 GiB at fp32).
3. Ran a full 120-problem eval; cosine-similarity scores at the 0.7 threshold matched the fp32 baseline.

**Memory + perf — JVM heap:**
1. Ran a load test harness against the local search-server, replaying ~1800 real `find_product` + `view_product_information` calls extracted from a prod-shaped eval trajectory.
2. Concurrency = 60 (matches `SANDBOX_MAX_WORKERS=60`), duration = 60 s per pass.
3. Three configurations tested (same harness, same query distribution, JVM cold-start cleared between passes):

| Heap | Throughput | p50 ms | p95 ms | p99 ms | Errors |
|---|---|---|---|---|---|
| Xmx4g (explicit) | 21.3 rps | 2722 | 5579 | 7118 | 0 / 1309 |
| Xmx3g | 17.4 rps | 3246 | 6774 | 7771 | 0 / 1146 |
| Xmx2g | 17.5 rps | 3237 | 7464 | 9437 | 0 / 1114 |

**Test Results:**

[...]

PR #205, opened 2026-06-28
scoring: opt-in shadow cosine logger + revert canonical to Qwen3 (ORO-1474)

that-guy-wade:
## Description

The v1.1.12 bge-small swap regressed the prod scoring distribution ~17pp before we caught it via miner reports. Pre-deploy validation sampled three bottom-band agents on a fresh staging suite; that under-represented the "almost-correct" rate top miners hit on the prod suite, which is exactly where bge@0.95 disagrees with Qwen3@0.7. Stable was rolled back to v1.1.11.

Adds a minimal observability hook so the next attempt is measured against the actual prod scoring distribution before any swap.

## Changes Made

- \`src/agent/rewards/orm.py\`:
  - Canonical model + threshold reverted to \`Qwen/Qwen3-Embedding-0.6B\` + 0.7 so the in-tree code matches what the deployed \`v1.1.11\` stable image runs.
  - When \`SHADOW_SCORE_MODEL\` is set, the title-similarity loop encodes the same \`(product_title, gt_title)\` pair with the named candidate model and emits one JSON line (\`shadow_sim\` prefix) carrying both cosines + both model names + both titles. Threshold is never enforced by the shadow path — analysis applies thresholds offline against the raw similarity column.
  - No DB write, no metric publish, no behaviour change to canonical scoring.
- \`docker/validator/Dockerfile\`: pre-cache both Qwen3 and bge-small so enabling the shadow at runtime doesn't trigger an unexpected HF download.
- \`tests/test_scoring_perf.py\`: positive + negative coverage of the shadow log path; existing canonical tests unchanged.

## Issue Link

- Related to: ORO-1474
- Closes: N/A — observability piece; the actual re-attempt of the swap is gated on the decision criteria spelled out in the Linear ticket (one week of data, top-20 agent canonical-vs-shadow delta within run-to-run noise).

### Manual Testing

Unit suite. End-to-end manual test is the operational rollout itself — set \`SHADOW_SCORE_MODEL=BAAI/bge-small-en-v1.5\` on the official ORO validator after this lands, watch CW Logs for \`shadow_sim\` lines, run \`filter @message like /shadow_sim/\` in Insights to confirm pair-level emission and absence of any canonical regression.

**Test Results:** 6/6 tests pass locally including two new shadow-path cases.

### Automated Testing

**Test Command(s):**

\`\`\`bash
.venv/bin/python -m pytest tests/test_scoring_perf.py -v
\`\`\`

## Documentation

- [ ] README updated

**Documentation Changes:** Inline docstrings cover the env-var contract + the per-pair log shape so a future replay script can rely on the field names.

## Checklist

- [x] Any dependent changes have been published and merged — none required; this is the dependent change for the data collection.

## Additional Notes

The shadow log is opt-in via env var so external validator operators see no behaviour change. The Dockerfile pre-cache costs ~150MB image-size for bge-small + ~600MB for Qwen3 (Qwen3 was already preloaded before the bge swap); no runtime cost when the env var is unset.

### DIFF
--- a/docker/validator/Dockerfile
+++ b/docker/validator/Dockerfile
@@ -19,9 +19,11 @@ WORKDIR /app
 COPY docker/validator/pyproject.toml docker/validator/uv.lock /app/
 RUN uv sync --frozen --no-dev --no-install-project
 
-# Pre-cache SentenceTransformer model (avoids runtime download)
+# Pre-cache SentenceTransformer models (avoids runtime download). The
+# canonical Qwen3 model is always loaded; the bge-small bundle is for the
+# optional ORO-1474 shadow scorer enabled via SHADOW_SCORE_MODEL at runtime.
 ARG HF_TOKEN=""
-RUN HF_HUB_ENABLE_HF_TRANSFER=0 HF_TOKEN=${HF_TOKEN} .venv/bin/python -c "from sentence_transformers import SentenceTransformer; SentenceTransformer('BAAI/bge-small-en-v1.5')"
+RUN HF_HUB_ENABLE_HF_TRANSFER=0 HF_TOKEN=${HF_TOKEN} .venv/bin/python -c "from sentence_transformers import SentenceTransformer; SentenceTransformer('Qwen/Qwen3-Embedding-0.6B'); SentenceTransformer('BAAI/bge-small-en-v1.5')"
 
 # Copy validator code
 COPY subnet/ /app/subnet/
--- a/src/agent/rewards/orm.py
+++ b/src/agent/rewards/orm.py
@@ -1,10 +1,24 @@
+import json
+import logging
+import os
 from collections import Counter
 
-SENTENCE_MODEL_NAME = "BAAI/bge-small-en-v1.5"
-TITLE_SIM_THRESHOLD = 0.95
+SENTENCE_MODEL_NAME = "Qwen/Qwen3-Embedding-0.6B"
+TITLE_SIM_THRESHOLD = 0.7
+
+# ORO-1474 shadow scoring: when SHADOW_SCORE_MODEL is set, encode the same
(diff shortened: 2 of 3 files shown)


shardi-b (approved):
Approved — no blocking issues. The canonical revert is faithful to v1.1.11 (Qwen3-Embedding-0.6B + bfloat16 + threshold 0.7, confirmed against #194/#203 history), and the shadow path is well-isolated for the documented bge-small rollout (pre-cached, 33 MB, loads from local cache).

Three non-blocking recommendations for follow-up:

1. **Contain shadow-load failures** (`orm.py` `_log_shadow_sim`): `_get_shadow_model()` is called outside the `try/except`, and `_get_shadow_model` only catches `(ImportError, OSError, ValueError)`. A load failure outside that set (e.g. `MemoryError`/torch `RuntimeError` when loading a second model on the memory-tight validator, per #194) would propagate through `rule_score_reward`, fail that problem's scoring future, and cost it its canonical score — violating the "no canonical impact" guarantee. Move the `_get_shadow_model()` call inside the `try/except`, or broaden the catch to `except Exception`. Low risk for bge-small specifically, but matters for the larger candidate models the next swap attempt will use.

2. **Lock the shadow lazy-init** (`_get_shadow_model`): scoring runs across `ThreadPoolExecutor` workers, so the unlocked `_shadow_model` global can be constructed by several threads at once on the first batch — an N× transient memory spike that is the most likely trigger for (1) on the constrained host.

3. **Avoid redundant shadow encodes**: `_log_shadow_sim` re-encodes the constant `product["title"]` every title iteration and issues two unbatched `encode()` calls per pair (2K passes vs. K+1 for K titles). Minor, opt-in only.

None of these block the merge; flagging for a quick follow-up before the actual model-swap re-attempt enables the shadow with a larger candidate model.

PR #212, opened 2026-07-01
validator: re-attach per-problem score to end-of-run progress flush

that-guy-wade:
## Description

`ProgressBatcher.batch_report` silently swallows transient backend errors (`progress_batcher.py:108`) and does not retry. Fast evals (~60–100s sandbox) finish scoring in one burst, so the batcher's periodic ticks never fire — the whole run hits the batcher exactly once at completion. If that single request drops, every per-problem `score` is lost. The subsequent `_upload_logs` progress call only carried `{status, logs_s3_key}`, so nothing repaired the null `score` field.

Frontend consensus grid (`consensus-grid.tsx:28`) colors cells green iff `score === 1.0`. Runs affected by the drop rendered as almost-all-wrong even when 25/30 problems succeeded — the run-level score was still 0.83 because backend's `compute_score_from_progress` falls back to status-based scoring (`validator.py:791`), which also emits the observed `WARNING Backend scoring: 29/30 problems missing score field` log line.

## Changes Made

- Refactor `ProblemProgressUpdate` construction into a shared `problem_result_to_update(result, logs_s3_key=UNSET)` helper in `progress_batcher.py`.
- `ProgressBatcher.batch_report` uses the helper.
- `_upload_logs` uses the helper for scored problems; falls back to the old status-only shape only when `progress_reporter.get_result(pid)` returns `None`.
- Add `ProgressReporter.get_result` accessor so main.py can pull the full `ProblemResult` when building the flush payload.
- Backend upsert is already idempotent (`validator.py:1365`), so the redundant score writes across both paths are safe.

## Issue Link

- Related to: N/A (bug found via prod diagnosis on 2026-06-30/07-01)
- Closes: N/A

### Manual Testing
Verified the null-score pattern in prod for two affected agents (Kaizer, DE_Kaizer). Confirmed the backend fallback warning in CloudWatch at the exact `complete_run` timestamps for both runs. Confirmed via direct DB query that `problem_progress` JSONB has `status` + `logs_s3_key` but no `score` on 29/30 problems for one representative run.

**Test Results:**
- 4 fallback warnings in the last 24 h across 2 agents, all with sandbox durations under 105 s.
- With the fix, `_upload_logs` sends the same shape the batcher sends periodically, so at least one of the two paths must succeed for `score` to land.

### Automated Testing

- 6 new tests for `problem_result_to_update` cover: score passthrough, zero-score preserved, `logs_s3_key` default UNSET vs passed-through, reasoning-summary gating on `reasoning_score`, inference counts gated on `inference_total`.
- 2 new tests for `ProgressReporter.get_result` cover populated and missing results.
- Full existing validator unit suite still passes (107 tests) — the batcher refactor is behavior-preserving.

**Test Command(s):**
```bash
uv run pytest subnet/tests/validator/test_progress_batcher.py subnet/tests/validator/test_progress_reporter_envelope.py -v
uv run pytest subnet/tests/validator/ -q --ignore=subnet/tests/validator/test_validator_integration.py
```

## Documentation

- [ ] README updated

**Documentation Changes:**
Inline docstring on `problem_result_to_update` explains why both paths exist and the idempotency assumption. Inline block comment in `_upload_logs` calls out the intent.

## Checklist

## Additional Notes

There is a pre-existing latent bug in the failure branch: `main.py:1117` calls `self.retry_queue.add_progress(...)`, but `RetryQueue` has no `add_progress` method. It would `AttributeError` if `report_progress` failed. Not touched in this PR; will file a separate ticket.

### DIFF
--- a/subnet/tests/validator/test_progress_batcher.py
+++ b/subnet/tests/validator/test_progress_batcher.py
@@ -1,0 +1,73 @@
+"""Tests for progress_batcher helpers."""
+
+from uuid import uuid4
+
+from oro_sdk.models import ProblemStatus
+from oro_sdk.types import UNSET
+
+from validator.progress_batcher import problem_result_to_update
+from validator.types import ProblemResult
+
+
+def _make_result(**overrides) -> ProblemResult:
+    defaults = dict(
+        problem_id=str(uuid4()),
+        category="product",
+        status=ProblemStatus.SUCCESS,
+        score=1.0,
+        score_dict={},
+        inference_failures=0,
+        inference_total=0,
+    )
+    defaults.update(overrides)
+    return ProblemResult(**defaults)
+
+
+class TestProblemResultToUpdate:
+    def test_carries_score_and_status(self):
+        r = _make_result(score=1.0, status=ProblemStatus.SUCCESS)
+        u = problem_result_to_update(r)
+        assert u.score == 1.0
+        assert u.status == ProblemStatus.SUCCESS
+
+    def test_zero_score_preserved(self):
+        r = _make_result(score=0.0, status=ProblemStatus.FAILED)
+        u = problem_result_to_update(r)
+        assert u.score == 0.0
+        assert u.status == ProblemStatus.FAILED
+
+    def test_logs_s3_key_default_unset(self):
+        r = _make_result()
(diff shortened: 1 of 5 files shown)


shardi-b (approved):

PR #214, opened 2026-07-03
sandbox: cap per-container stdout at 10MB via docker --log-opt max-size

that-guy-wade:
## Description

Race 80 (2026-07-02) shipped **78 GB of CW ingest in a single day** (baseline ~2.5 GB) from a runaway agent stuck in a tight print loop in its LLM-error retry path. The two stderr lines below repeated at microsecond spacing across the race window:

```
# error parsing LLM response for decomposition: object of type 'NoneType' has no len()
# trigger_llm: global deadline reached, returning safe empty value
```

The agent's `trigger_llm()` returns `None` when the deadline elapses; the decomposition parser calls `len(None)` and raises; the error handler retries `trigger_llm()` which returns `None` again; loop. The pattern shows up in the London-family agent template.

`SANDBOX_LOG_MAX_BYTES=256 MiB` per-problem cap is too loose to keep this out of the daemon's log driver, and the awslogs driver on prod ships every rotation to CloudWatch. Cost impact: ~$41 on the one day, projected ~$1200/mo if the pattern recurs across future races.

Fix: apply `--log-opt max-size=10m --log-opt max-file=1` at `docker run` time inside `build_sandbox_command`. Typical evals write ~150 KB of stdout, so 10 MB gives ~66x headroom. Applies to whichever log driver the host is configured with (awslogs in prod, json-file locally).

**Zero eval impact.** Trajectory data flows through the sandbox `output.jsonl` envelope, not stdout — so scoring, reasoning judge, and S3 upload are all unaffected.

## Changes Made

- `subnet/sandbox.py`: add `--log-opt max-size=10m` + `--log-opt max-file=1` to the `docker run` command built by `build_sandbox_command`.
- `subnet/tests/test_sandbox.py`: new `test_log_size_capped` asserts both flags are present so a future edit can't silently reintroduce rotation-based inflation.

## Issue Link

- Related to: N/A (found via CW billing audit 2026-07-03)
- Closes: N/A

### Manual Testing
Confirmed the flood pattern in prod CW: `/oro/validator/production` on 2026-07-02 = 78.76 GB vs 2026-06-30 = 3.81 GB (same 60-problem race count). Sampled the burst hosts and found the two-line stderr loop at microsecond spacing.

**Test Results:**
Not applicable — no runtime code path changed for good agents; only the log driver cap.

### Automated Testing

**Test Command(s):**
```bash
uv run pytest subnet/tests/test_sandbox.py -v
```
35 tests pass including the new `test_log_size_capped`.

## Documentation

- [ ] README updated

**Documentation Changes:**
Inline comment on the docker command explains why the cap exists and cites the race 80 incident.

## Checklist

## Additional Notes

Follow-ups (separate tickets):
- Agent template hardening — decomposition error-handler should give up after N failures instead of tight-looping.
- Nginx proxy `access_log off` inside `/search/` block — orthogonal noise-reduction, but the sandbox stdout cap is the higher-leverage fix.

### DIFF
--- a/subnet/sandbox.py
+++ b/subnet/sandbox.py
@@ -158,12 +158,17 @@ def build_sandbox_command(
 
     cmd.extend(
         [
-            # Cap per-container stdout so a runaway agent (e.g. tight print loop
-            # in an error-handler retry path) cannot flood the host log driver.
-            # Typical eval writes ~150 KB, so 10 MB is a ~66x headroom while
-            # killing the "1 MB/s microsecond spin" pattern that shipped 78 GB
-            # of CW ingest during race 80. Applies to whichever log driver the
-            # host is configured with (awslogs in prod, json-file locally).
+            # Force json-file with a hard 10 MB cap. The prod daemon default is
+            # awslogs, which rejects max-size / max-file at container-create
+            # time (would break every eval), and which shipped 78 GB of CW
+            # ingest during race 80 when a runaway agent spun in a tight
+            # print loop in its LLM-error retry path. Sandbox stdout is
+            # already redundant with trajectory JSON (in S3) and the
+            # sandbox_metadata.stderr_tail persisted on eval_run, so the
+            # json-file sink is only needed for live SSH `docker logs` debug.
+            # Typical eval writes ~150 KB → 10 MB is ~66x headroom.
+            "--log-driver",
+            "json-file",
             "--log-opt",
             "max-size=10m",
             "--log-opt",
--- a/subnet/sandbox.py
+++ b/subnet/sandbox.py
@@ -156,38 +156,50 @@ def build_sandbox_command(
     if container_name:
         cmd.extend(["--name", container_name])
 
-    cmd.extend([
-        # Resource limits — prevent runaway miner agents from impacting the host
-        "--memory",
-        "4g",
-        "--memory-swap",
-        "4g",
-        "--pids-limit",
-        "256",
-        "--ulimit",
-        # 4096 fds: at SANDBOX_MAX_WORKERS=60 a typical agent holds ~5 sockets
-        # in flight per worker (HTTP + DNS + keepalive), pushing toward the
-        # 1024 cap and silently capping throughput before the worker count.
-        "nofile=4096:4096",
-        # CPU priority — sandbox runs during the validator's blocked
-        # subprocess.run(); only the heartbeat + weight-setter threads compete
-        # for CPU at that point, so giving sandbox 2x the default share keeps
-        # those threads responsive while no longer starving sandbox workers on
-        # smaller (8 vCPU) validator hosts at SANDBOX_MAX_WORKERS=60.
-        "--cpu-shares",
-        "2048",
-        "--user",
-        "1000:1000",
-        # Security hardening — minimize container attack surface
-        "--cap-drop=ALL",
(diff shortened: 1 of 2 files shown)


shardi-b (dismissed):

## Critical (correctness): `--log-opt max-size/max-file` are incompatible with the `awslogs` driver this PR targets

The PR adds `--log-opt max-size=10m --log-opt max-file=1` to `docker run` but does **not** set `--log-driver`, so the options are validated against the daemon default driver. The PR body states that default is **awslogs in prod**.

`max-size`/`max-file` are options of the **json-file / local** drivers only. The `awslogs` driver does not accept them, and Docker validates log options per-driver **at container-create time**, erroring rather than ignoring:

```
docker: Error response from daemon: unknown log opt 'max-size' for awslogs log driver.
```

Consequences on a prod host whose default driver is awslogs:
- Every `docker run` from `build_sandbox_command` fails at launch → **all evaluations fail** on that validator.
- Even if the option were tolerated, `max-size`/`max-file` govern local json-file rotation, not the awslogs→CloudWatch stream — so the **78 GB CloudWatch ingest this PR aims to cap would not be reduced**.

This works only on a json-file-default host (local dev), which is exactly where CloudWatch is not involved. It passes locally/CI and fails in the one environment it was written for.

### Suggested fix — set the driver explicitly so behavior is host-independent
- `--log-driver json-file --log-opt max-size=10m --log-opt max-file=1` — guarantees the cap applies on every host, **or**
- `--log-driver none` — the validator already captures the container's stdout/stderr via the attached pipe in `subnet/validator/bounded_io.py` (capped at `SANDBOX_LOG_MAX_BYTES`, 256 MiB), so the daemon log sink may be unnecessary; this removes CloudWatch ingest for sandbox containers entirely.

## Secondary (test quality): `test_log_size_capped` is presence-only and green-lights the broken config

`assert "max-size=10m" in cmd` matches the token anywhere in the list, decoupled from its `--log-opt` and independent of the effective driver. Since the CI/local default is json-file, the test stays green while the prod (awslogs) path fails — it structurally cannot catch the real defect. The test should assert an explicit `--log-driver` is present and that each value immediately follows its `--log-opt`.

Refs: `subnet/sandbox.py:159-190`, `subnet/tests/test_sandbox.py:130-139`, `subnet/validator/bounded_io.py`.

that-guy-wade:
Good catch on both points — the awslogs incompatibility would have failed every eval at container-create, and the presence-only test would not have caught it.

Fixed in `f31cd9d`:

- Added `--log-driver json-file` immediately before the two `--log-opt` flags. This decouples the cap from the daemon default, so it applies uniformly on prod (awslogs-default) hosts and local dev (json-file-default) hosts.
- Chose `json-file` over `none`. Sandbox stdout is already redundant with the trajectory JSON in S3 (parsed dialogue) and the `sandbox_metadata.stderr_tail` persisted on `eval_run`, so we don't lose anything eval-critical, but `docker logs <container>` still works for live SSH debug during a stuck eval.

Test rewritten to:
- Assert `--log-driver json-file` is present.
- Walk each `--log-opt` and check its value sits directly after (`cmd[i+1]`) rather than "somewhere in the list". A future edit that drops `--log-driver` or reorders the flags will now fail the test rather than silently regressing.

shardi-b (approved):

PR #215, opened 2026-07-07
search-server: two-tier BM25 candidate cap (k=500 → k=10000)

that-guy-wade:
## Description

Cuts search-server CPU by scoring only 500 BM25 candidates on the first
attempt instead of 10,000. Falls back to the full 10,000 candidate cap
when the post-filter (shop_id / price / service) yield is short.

The searcher was running `LuceneSearcher.search(k=10000)` on every
request even though the endpoint returns at most 50 products across
five pages. Most queries only need the top few hundred to fill the
response.

## Changes Made

- `src/search_engine/server.py`
  - Split \`CAPACITY\` into \`CAPACITY_INITIAL = 500\` +
    \`CAPACITY_FULL = 10000\`.
  - Extract the per-attempt logic into a local \`_collect(k)\` helper
    that runs the BM25 search + post-filters, returning
    \`(products, len(hits))\` so the caller can distinguish
    "corpus ran out" from "post-filter dropped everything".
  - Two-tier fetch: try \`k=CAPACITY_INITIAL\`; escalate to
    \`k=CAPACITY_FULL\` only when both conditions hold —
    (a) the post-filter yield is < \`TARGET_HITS\` (50), AND
    (b) the searcher actually still has more candidates to score.

### Manual Testing
Ran a regression harness against two identical Docker instances of the
search server (baseline flat \`k=10000\` vs candidate two-tier), driving
several thousand real \`find_product\` parameter tuples through both
under concurrency=24 and diffing top-50 product-id sequences.

**Test Results:**

| variant                 | p50    | p95     | p99     | max     | correctness |
|-------------------------|--------|---------|---------|---------|-------------|
| baseline k=10000        | 2256ms | 14674ms | 27283ms | 29934ms | ref         |
| candidate k=500→10000   |   64ms |  1181ms |  1935ms |  2279ms | 100% match  |

Zero reordering, zero content divergence on the successful queries.

### Automated Testing
Existing unit tests pass locally against the modified server.

**Test Command(s):**
\`\`\`bash
docker run -d --name a -p 5632:5632 <image>       # baseline
docker run -d --name b -p 5633:5632 -v .../server.py:/app/src/search_engine/server.py <image>
python3 regression.py
\`\`\`

## Documentation

- [ ] README updated

**Documentation Changes:**
Inline comment on the constants explains the two-tier strategy and
cites the regression finding without pointing at any internal ticket.

## Checklist

## Additional Notes

Deploy plan: Watchtower on validator hosts picks up the new image on
the next promote. No config change required — the new constants are
in-source.

Follow-ups (larger, separate PRs):
- Push shop_id into Lucene as a keyword field so filtered queries can
  set \`k = TARGET_HITS\` directly and avoid post-filter altogether.
- LRU cache on \`(q, filters)\` to catch repeat calls from a single
  agent trajectory.

### DIFF
--- a/src/search_engine/server.py
+++ b/src/search_engine/server.py
@@ -51,9 +51,16 @@ def sanitize_query(q: str) -> str:
     return q.translate(_LUCENE_SPECIAL_CHARS).strip()
 
 
-CAPACITY = 10000
+# Two-tier BM25 candidate cap: try the cheap tier first, fall back to the
+# full cap only when the post-filter (shop_id / price / service) yield is
+# short. A regression on thousands of real find_product tuples confirmed
+# 100% top-50 identity vs the flat k=10000 path, with a large median and
+# tail latency win under concurrency.
+CAPACITY_INITIAL = 500
+CAPACITY_FULL = 10000
 PAGE_SIZE = 10
 MAX_PAGE = 5
+TARGET_HITS = PAGE_SIZE * MAX_PAGE  # final products returned across pages
 SEARCH_FIELDS = ["product_id", "shop_id", "title", "price", "service", "sold_count"]
 INFORMATION_FIELDS = [
     "product_id",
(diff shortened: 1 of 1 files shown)


shardi-b (approved):

PR #217, opened 2026-07-08
fix(validator): retry inference-token smoke-test on transient 401

that-guy-wade:
## Description

Before running an eval, the validator smoke-tests the miner's freshly-minted scoped inference token with a 1-token completion (`_validate_inference_token`). A newly-created provider key can return **401 briefly while the provider propagates it** — the token is valid and works moments later. The smoke test treated that 401 as a hard failure and failed the run immediately, even though the same token (or a re-claimed one) succeeds shortly after.

The function already treats 5xx / 429 / timeout as transient (proceeds without failing); this extends the same treatment to a propagation-lag 401 by retrying before giving up.

## Changes Made

- `subnet/validator/main.py`: `_validate_inference_token` now retries a 401 up to 3 times with a 1.5s backoff before returning invalid. 200 short-circuits to valid; 402 (no credits) and other statuses are unchanged. A genuinely invalid/revoked key stays 401 through every attempt and fails fast (~4.5s worst case), after which the run re-claims a fresh token as before.
- `subnet/tests/validator/test_validate_inference_token.py`: updated the persistent-401 test to assert it retries then fails (4 attempts, 3 sleeps, `time.sleep` patched); added a 401-then-200 test asserting the retry recovers to valid.

### Manual Testing

`_validate_inference_token` unit tests: 13 passed. Broader validator suite: 184 passed.

**Test Command(s):**
```bash
pytest subnet/tests/validator/test_validate_inference_token.py -q
pytest subnet/tests/validator/ -q
```

## Additional Notes

Only the 401 path loops; happy-path latency is unchanged. Worst-case added latency is ~4.5s, and only on runs whose token initially 401s (rare). No config surface; retry count/backoff are local constants.

### DIFF
--- a/subnet/tests/validator/test_validate_inference_token.py
+++ b/subnet/tests/validator/test_validate_inference_token.py
@@ -52,13 +52,38 @@ class TestValidateInferenceToken:
         assert ok is True
         assert reason == ""
 
-    def test_401_returns_invalid(self):
-        with patch("validator.main.requests.post", return_value=self._mock_resp(401)):
+    def test_401_persists_fails_after_retries(self):
+        # A genuinely-bad key stays 401 through every attempt → fails fast.
+        with (
+            patch("validator.main.time.sleep") as mock_sleep,
+            patch(
+                "validator.main.requests.post", return_value=self._mock_resp(401)
+            ) as mock_post,
+        ):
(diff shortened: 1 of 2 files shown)


shardi-b (approved):

PR #218, opened 2026-07-09
fix(validator): protect the survivor set, not top-50% of finishers

that-guy-wade:
## Description

The validator's deregistration-protection tail was defined as the **top 50% of last race's finishers**, independent of the elimination fraction (the bottom-cut applied at race end). With today's 65% elimination cut, finishers ranking between the survival line (top 35%) and the 50% tail line were **eliminated yet still received emission weight**. Two thresholds — the elimination fraction and the fixed 50% tail — had to be kept in sync by hand, and nothing enforced it.

This ties the protected tail to the **survivor set** instead: an agent is protected iff it survived the race (finished, not eliminated, not discarded) and advances to the next one. The protected set now tracks the elimination fraction automatically — no second threshold to maintain.

No backend or SDK change is required: the public `RaceQualifierPublic` record already exposes `eliminated_at`.

## Changes Made

- `_qualifiers_to_finishers`: drop `eliminated_at` qualifiers alongside the existing `is_discarded` drop, so `finishers` is the survivor set.
- `compute_hotkey_weights`: distribute the linear taper across **all survivors** rather than the top 50% (removed the `len(ranked) // 2` cut).
- Updated module/function docstrings from "top-50%" to the survivor-set contract.
- Migrated `test_weight_distribution.py` and `test_weight_setter.py` to the survivor-set contract; added a `TestQualifiersToFinishersEliminated` class covering the eliminated drop (including forward-compat for missing/UNSET fields). Fixed the `_race_detail` mock helper to set `eliminated_at` explicitly — a bare `MagicMock` returns a truthy child for unset attributes, which would make every qualifier look eliminated.

## Issue Link

- Related to: N/A
- Closes: N/A

### Manual Testing

Backtested the new logic against the most recent completed race (399 finishers: 258 eliminated, 141 survivors) on the live metagraph, and validated the harness by recomputing the **old** logic and confirming it reproduces the current on-chain vector exactly.

**Test Results:**

Harness validation — old logic recomputed from race finishers reproduces the live on-chain validator vector byte-for-byte:
```
on-chain:      top 85.52%   102 nonzero
old-recomputed top 85.52%   102 nonzero   ✓ match
```

Effect of the fix (`emission_baseline_burn_rate = 0.0`, current prod):
```
                 top miner   protected tail   -> eliminated agents
OLD (top-50%)     85.52%       101 entries          0.878%
NEW (survivor)    91.59%        76 entries          0.000%
```

Decomposition of the tail change (top gain = tail loss):
```
OLD tail = 14.484%  = survivors 13.606% + eliminated 0.878%
NEW tail =  8.412%  = survivors  8.412% + eliminated 0.000%
top gain = +6.072%  = survivors-reduced 5.194% + eliminated-removed 0.878%
```

Notes on the +6% top-miner increase (reviewed and accepted): with `t_burn = 0` there is no burn slot to absorb the freed tail budget, so the top miner (pinned at `U16_MAX`) takes a larger normalized share when the tail shrinks. ~5.2% of the shift is survivors receiving smaller taper values (the taper scales to tail size `M`, which drops from ~199 to ~140); 0.878% is the removed eliminated leak. Under any `t_burn > 0` the top stays pinned at exactly `t_top` and the freed share flows to burn instead — the "top share invariant under N" property still holds (covered by `test_top_miner_share_is_exactly_t_top_regardless_of_n`).

Sample NEW chain vector (u16 = submitted weights):
```
uid   u16     share    role
 25  65535   91.59%    top
204    140    0.20%    survivor
146    139    0.19%    survivor
...    (taper 140→1 across 76 registered survivors)
rest     0             not protected
```

### Automated Testing

**Test Command(s):**
```bash
python -m pytest subnet/tests/validator/test_weight_setter.py subnet/tests/validator/test_weight_distribution.py -q
# 78 passed
```

## Documentation

- [ ] README updated

**Documentation Changes:** Module and function docstrings updated to describe the survivor-set contract and why it decouples protection from the elimination fraction.

## Checklist

## Additional Notes

Determinism (byte-identical vectors across validators) is preserved — the survivor set is deterministic shared state (`eliminated_at` is the same for every validator), so the property is as strong as before. This changes on-chain emission behavior, so it warrants careful review before it ships.

### DIFF
--- a/subnet/tests/validator/test_weight_distribution.py
+++ b/subnet/tests/validator/test_weight_distribution.py
@@ -1,4 +1,4 @@
-"""Tests for the deterministic top-50% weight distribution.
+"""Tests for the deterministic survivor-set weight distribution.
 
 Determinism is load-bearing for Yuma consensus on subnet 15
 (`kappa = 0.5`): if two validators emit different weight vectors for the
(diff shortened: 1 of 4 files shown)


shardi-b (approved):

PR #220, opened 2026-07-09
fix(grader): normalize attribute matching + same-key token-subset

that-guy-wade:
## Description

The per-problem grader (`src/agent/rewards/orm.py:rule_score_reward`) matched product attributes/sku to the reward by **exact `(key, value)` tuple** with no normalization, and binary success requires every constraint hit. This produced systematic **false-negatives**: a product that genuinely satisfies the query is scored incorrect when it encodes an attribute under a different key spelling or with extra descriptor tokens than the reference target.

Examples (both were valid products scored 0):
- reward `size: "48mm"` vs product `size: "1pcs 2#- 48mm"` (same size, extra packaging tokens)
- reward `color family: "..."` vs product `color_family: "..."` (space vs underscore)
- reward `color_family: "7*9cm 10pcs"` vs product `color_family: "7*9cm【10pcs】"` (full-width brackets)

Title (semantic embedding) and price (numeric range) matching are already robust — only attribute/sku matching is affected.

## Changes Made

- Replaced the exact `(key, value)` attribute/sku match with a layered matcher, all **exact-after-normalize** (no fuzzy/embedding):
  - **Value normalization** — NFKC + lowercase + strip brackets/full-width/whitespace; values still compared for exact equality, so distinct enums never collide (`43`≠`42`, `blue`≠`black`, `2023`≠`2024`).
  - **Key string-normalization** — space/underscore/case/plural equivalence (`color family`==`color_family`, `flavors`==`flavor`).
  - **Same-key token-subset** — a reward value matches if its tokens are a whole-token subsequence of one of the product's values under that (normalized) key (`48mm` in `1pcs 2#- 48mm`). Whole-token boundaries keep `8gb`≠`128gb`.
- Added `tests/test_attr_matching.py`.

Semantic cross-key aliases (e.g. `color`↔`color_family`, `model`↔`brand`) are **intentionally out of scope** here — they require corpus-derived key co-occurrence stats (with per-suite determinism handling) and are a planned follow-up.

## Issue Link

- Related to: N/A
- Closes: N/A

### Manual Testing

Backtested against a full historical race with the real scorer + live product catalog: re-scored sampled score-0 and score-1 product evals and compared to the recorded verdict.

**Test Results:**
```
score-0 sample (n=381): recovered (now pass) = 46  (12.1%)
score-1 sample (n=400): regressions (now fail) = 0  (0.0%)
```
Zero regressions is expected structurally — the matcher is strictly more permissive than exact match (normalization only adds matches), so no previously-passing eval can break. Recovered cases are token-subset (`48mm` in `1pcs 2#- 48mm`) and key-spelling variants.

### Automated Testing

**Test Command(s):**
```bash
python -m pytest tests/test_attr_matching.py -v
# 11 passed
```
Covers value/key normalization, token-subsequence boundaries, positive matches (key-variant, bracket, token-subset), and the negative invariants (`blue`≠`black`, `8gb`≠`128gb`, `2023`≠`2024`, `v1 red lining`≠`v1 black lining`, product-underspecifies `yellow` vs `yellow-24inch`, and that semantic cross-key is *not* matched here).

## Documentation

- [ ] README updated

**Documentation Changes:** Inline docstrings on the new matcher helpers explaining the exact-after-normalize invariant and the deliberate exclusion of semantic cross-key aliasing.

## Checklist

## Additional Notes

This changes on-chain scoring, so it should roll out fleet-wide before it takes effect — mixed grader versions across validators would score the same agent differently. The matcher is fully deterministic (pure string ops), so validators on the same image produce identical results. Applies forward.

### DIFF
--- a/src/agent/rewards/orm.py
+++ b/src/agent/rewards/orm.py
@@ -1,6 +1,8 @@
 import json
 import logging
 import os
+import re
+import unicodedata
 from collections import Counter
 
 SENTENCE_MODEL_NAME = "BAAI/bge-small-en-v1.5"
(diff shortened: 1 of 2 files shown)


shardi-b (approved):

Reviewed with high-effort multi-angle analysis (correctness, over-matching/gaming, removed-behavior, cross-file, cleanup/altitude, test correctness, conventions).

Structural safety: the new matcher is strictly monotonic (normalization + token-subset only add matches), so hit_count can only rise and no previously-passing eval can regress downward — consistent with the backtest (0 regressions on n=400). Removed product guards are correctly re-established via product.get(...); the return tuple and total_counter/hit_counter keys consumed by problem_scorer.py are unchanged.

The potential adversarial gaming vector (token-stuffing a single attribute value to satisfy many same-key constraints) was investigated and REFUTED: product attributes are fetched from the search-server catalog via get_product(); miners submit only product IDs, so attribute strings are not attacker-controlled.

Non-blocking observations for a possible follow-up (none block this PR):
- Same-key token-subset can credit near-miss variants (e.g. reward "large" matches product "extra large", "red" matches "dark red"). This is the intended, scoped, backtested tradeoff.
- normalize_attr_key strips a trailing "s" without a plurality check, mangling non-plural keys (lens->len, class->clas); only additive, low collision probability.
- _build_attr_index(attr_pairs + option_pairs) re-indexes the invariant attribute set on every sku-option iteration; could build the attr base once outside the loop.
- Tests cover the helpers thoroughly but do not exercise rule_score_reward end-to-end, and the subsequence ordering semantics are not asserted.

PR #222, opened 2026-07-10
feat(grader): curated cross-key attribute aliases

that-guy-wade:
## Description

Adds a curated cross-key alias layer to the attribute matcher so a product that encodes a required attribute value under a *semantically-equivalent sibling key* satisfies the constraint. Example: a query requiring `color_family = "blue"` is satisfied by a product listing `color = "blue"`; a `model = "samsung galaxy a04e"` requirement is satisfied by `compatibility_by_model = "samsung galaxy a04e"`.

These are real false-negatives — the product genuinely meets the requirement, but the reward and the product happen to use different (equivalent) key names for the same slot.

The list is an **explicit hand-vetted allowlist**, not derived dynamically. A corpus co-occurrence approach was prototyped and rejected: the catalog is dirty (sellers dump compound junk like `"4 modes 12v"` into unrelated fields such as both `voltage` and `color`), so no automatic distinctiveness/support threshold separates real aliases from coincidence. The allowlist is also deterministic by construction — identical across all validators with no per-suite artifact to ship.

## Changes Made

- `src/agent/rewards/orm.py`
  - Added `_CROSS_KEY_ALIAS_PAIRS` (15 hand-vetted pairs) + normalized `_CROSS_KEY_ALIASES` frozenset.
  - Extended `_attr_constraint_hit` with a third match layer (after same-key exact + same-key token-subset): exact value found under a curated sibling key.
- `tests/test_attr_matching.py`
  - Added positive (allowlisted pairs match, incl. real reported cases), exact-value-only, and negative (non-allowlisted pairs like `voltage`/`color` never cross-match) tests. Updated the module docstring.

## Issue Link

- Related to: N/A
- Closes: N/A

### Manual Testing
Verified the reported false-negative products now reach attribute-full under the new matcher (necklace `colored_gem_type`↔`main_stone`, phone battery `model`↔`compatibility_by_model`), while distinct-value and non-allowlisted-key cases still correctly fail.

**Test Results:**
Offline backtest over ~213k historical evals: the curated list recovers ~126 distinct (problem, product) false-negatives (~8k eval-level) with the deliberately-excluded junk key-pairs (`color_family`↔`size`, `voltage`↔`color`, …) producing zero matches. `color`↔`color_family` alone accounts for ~60% of the recovery.

### Automated Testing
Attribute-matcher unit tests (existing + new cross-key cases).

**Test Command(s):**
```bash
python -m pytest tests/test_attr_matching.py -q
python -m ruff check src/agent/rewards/orm.py tests/test_attr_matching.py
```

## Documentation

- [ ] README updated

**Documentation Changes:**
Inline rationale on the allowlist explaining why it is explicit rather than corpus-derived.

## Checklist

## Additional Notes

This changes on-chain scoring (grader). **Not for immediate merge/deploy** — it should land fleet-wide, deterministically, and bundle into a grader release *after* the current attribute-normalization change has baked a race or two (no two grader changes in the same race window). Value-level matches within legit key-pairs are exact-after-normalize; the allowlist restricts the key dimension so unrelated-key over-matches cannot occur.

### DIFF
--- a/src/agent/rewards/orm.py
+++ b/src/agent/rewards/orm.py
@@ -86,12 +86,44 @@ def _build_attr_index(kv_pairs):
     return val_keys, key_tokens
 
 
+# Curated cross-key aliases: pairs of attribute keys that are semantically the
+# same slot, so the SAME value found under either key satisfies the constraint
+# (e.g. a `color_family` requirement met by the product's `color`). Hand-vetted:
+# a dynamic corpus-co-occurrence approach over-matched because the catalog is
+# dirty — unrelated keys share junk values (e.g. `voltage`/`color`
+# both carry "4 modes 12v"), and no threshold separates real aliases from
+# coincidence. So this is an explicit allowlist. Pairs are stored normalized.
+_CROSS_KEY_ALIAS_PAIRS = [
+    ("color", "color_family"),
+    ("color_classification", "color_family"),
+    ("color_classification", "sort by color"),
+    ("color_family", "warna"),  # 'warna' == color (id/ms)
+    ("plate_size", "size"),
+    ("cord_length", "length"),
+    ("cookware_material", "material"),
+    ("finish_lipstick", "lips makeup finish"),
+    ("colored_gem_type", "main_stone"),
+    ("model", "compatibility_by_model"),
+    ("compatibility", "compatibility_by_model"),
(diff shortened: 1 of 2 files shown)


shardi-b (approved):

PR #223, opened 2026-07-12
perf(search-server): multi-worker (gunicorn) to fix search queuing under concurrency

that-guy-wade:
## Description

Serve the search-server via multiple gunicorn worker **processes** instead of a single waitress process, to fix search-request queuing under concurrency on slow / CPU-contended validator hosts.

**Why:** the search endpoint's per-request cost is Lucene BM25 search (~60%) + per-doc raw fetch of the candidate set (~38%), both via JNI. pyserini/pyjnius holds the GIL across JNI calls, so a single process serializes requests once its GIL-bound throughput ceiling is hit — leaving cores idle. On a host with CPU steal (a stolen vCPU freezes the GIL-holding thread), this collapses to ~1 effective core and search latency balloons under the race concurrency burst (~64 concurrent searches), pushing problems past the 300s per-problem timeout.

Measured on a real validator (via trajectory `proxy_calls`): search latency is 18ms in isolation but ~4.7s at 64 concurrent in-flight searches on the affected host, vs ~49ms on a healthy host that parallelizes.

**Fix:** gunicorn with `SEARCH_WORKERS` processes, each its own GIL + JVM. The mmap'd Lucene index is shared across workers via the OS page cache (~1× index RAM). No `--preload` — the embedded JVM does not survive `fork()`, so each worker initializes post-fork.

## Changes Made

- `docker/search-server/entrypoint.sh` — launch gunicorn with `SEARCH_WORKERS` processes (default: `nproc/2`, floor 2; set `SEARCH_WORKERS=1` for single-process). Keeps the waitress `__main__` in server.py for local `python server.py` dev.
- `docker/search-server/Dockerfile` — install gunicorn; per-worker JVM heap reduced to `-Xmx1g` (index is off-heap/mmap, so on-heap need per worker is small).
- `docker/search-server/requirements.txt` — add `gunicorn==23.0.0`.

## Issue Link

- Related to: N/A
- Closes: N/A

### Manual Testing
Benchmarked waitress (single process) vs gunicorn 4-worker against the real `/find_product` (k=500) endpoint at the observed concurrency (C=64), on a 10-core box.

**Test Results:**

| rig (C=64) | waitress | gunicorn-4 |
|---|---|---|
| unthrottled | 84 req/s, p95 960ms | **120 req/s, p95 728ms** (1.4×) |
| 3-core noisy-neighbor (spare cores) | 53 req/s, p95 **1872ms** | **83 req/s, p95 931ms** (1.55×, ½ tail) |
| cores saturated | ~71 req/s | ~70 req/s (tie) |

Multi-worker wins whenever spare cores exist (the target condition); it ties when cores are saturated — hence the default caps workers at half the cores. Search results are identical (concurrency change only, same index).

### Automated Testing
Existing search-server behavior unchanged; the WSGI `app` and `LuceneSearcher` are unchanged — only the process model differs.

**Test Command(s):**
```bash
# concurrency benchmark against /find_product at C=64
```

## Documentation

- [ ] README updated

**Documentation Changes:**
Inline rationale in entrypoint.sh (why processes not threads; why workers ≤ cores; no --preload).

## Checklist

## Additional Notes

Infra/perf change only — does not touch scoring or the index (identical results). Intended primarily to be A/B'd on a slow/contended validator host (`SEARCH_WORKERS` tunable). Complementary follow-up: doc-values/query-time filtering to cut the 38% doc-fetch (bigger effort — needs a corpus re-index).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,15 +8,23 @@ services:
     environment:
       - HOST=${HOST:-0.0.0.0}
       - PORT=${PORT:-5632}
-      - WAITRESS_CONNECTION_LIMIT=${WAITRESS_CONNECTION_LIMIT:-1000}
-      # JVM heap passthrough — Dockerfile sets a default, override here per host.
-      - _JAVA_OPTIONS=${_JAVA_OPTIONS:--Xmx2g -Xms1g}
+      # gunicorn worker processes (default nproc/2, floor 2) x threads/worker.
+      # Keep workers <= free cores so the extra JVMs don't oversubscribe.
+      - SEARCH_WORKERS=${SEARCH_WORKERS:-}
+      - SEARCH_THREADS=${SEARCH_THREADS:-4}
+      # Per-worker JVM heap. Must match the Dockerfile default (this compose var
+      # overrides the image ENV). Index is off-heap/mmap, so 1g/worker is plenty;
+      # keeps total heap modest across N workers on memory-constrained hosts.
+      - _JAVA_OPTIONS=${_JAVA_OPTIONS:--Xmx1g -Xms256m}
     ports:
       - "${PORT:-5632}:${PORT:-5632}"
     healthcheck:
       test: ["CMD-SHELL", "python -c \"import os; import urllib.request; urllib.request.urlopen('http://localhost:' + os.environ.get('PORT', '5632') + '/health').read()\""]
       interval: 10s
-      timeout: 5s
+      # 10s (was 5s): under a heavy concurrent search burst a health probe may
+      # briefly wait for a gthread slot; a tighter timeout risks false unhealthy
+      # flaps on the contended hosts this change targets.
+      timeout: 10s
       retries: 3
       start_period: 30s
     labels:
--- a/docker/search-server/Dockerfile
+++ b/docker/search-server/Dockerfile
@@ -22,6 +22,11 @@ RUN set -eux; \
     apt-get autoremove -y; \
     rm -rf /var/lib/apt/lists/*
 
+# gunicorn is the production WSGI server (multi-process). Also in
+# requirements.txt for the base image; installed here too so the code-layer
+# image is self-contained even before the base is rebuilt.
+RUN pip install --no-cache-dir gunicorn==23.0.0
+
 # Only the code layer — changes on every release
 COPY src/search_engine /app/src/search_engine
 COPY src/__init__.py /app/src/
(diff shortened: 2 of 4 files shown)


shardi-b (dismissed):

## Findings

**1. (blocking) `docker-compose.yml:13` negates the per-worker heap reduction — OOM risk on the targeted hosts.**
The Dockerfile now sets `_JAVA_OPTIONS="-Xmx1g -Xms256m"` (per-worker), and the PR rationale ("4 workers -> ~4 GiB") depends on that 1g cap. But `docker-compose.yml` — the actual deploy path — sets `_JAVA_OPTIONS=${_JAVA_OPTIONS:--Xmx2g -Xms1g}`, and a compose `environment` var overrides the image ENV. So each of the N gunicorn workers runs a **2g** heap, not 1g. On a 10-core host the entrypoint defaults to `nproc/2 = 5` workers → ~10 GiB max heap (5×1g committed via Xms) instead of ~4 GiB. On the memory-constrained / CPU-steal hosts this PR explicitly targets, this risks OOM-killing workers. Please update the compose default to match (or drop the override).

**2. `entrypoint.sh` `--worker-class sync` can starve `/health` under the target burst.**
Sync workers serve exactly one request each. With ~5 workers all occupied by 0.7–4.7s searches during the ~64-concurrent burst, a `/health` request queues behind in-flight searches and can exceed the compose healthcheck `timeout: 5s`; after `retries: 3` the container flaps **unhealthy** — under the exact contended-host condition this PR targets, and where `depends_on: service_healthy`/watchtower react to it. Waitress ran 32+ threads so near-instant `/health` always had a slot. Consider `--worker-class gthread` with a few threads, a raised health timeout, or reserving capacity for health.

**3. `SEARCH_WORKERS` is not wired into `docker-compose.yml`.**
The PR documents `SEARCH_WORKERS` as the tuning knob (incl. `SEARCH_WORKERS=1` fallback), but the compose `search-server` `environment` block does not forward it, so on the primary deploy path it cannot be set without editing compose. Also `WAITRESS_CONNECTION_LIMIT` (compose line 11) is now injected but ignored by gunicorn — worth removing to avoid confusing operators.

**4. (minor) `--access-logfile -` adds per-request logging to the hot search path.**
At the ~120 req/s target burst this writes one access-log line per `/find_product` — pure overhead the previous waitress server did not incur. Consider disabling.

Verified as non-issues: the `src.search_engine.server:app` gunicorn target imports fine despite no `__init__.py` (namespace subpackage); `--chdir /app` matches the old WORKDIR so `indexes` still resolves; no-`--preload` post-fork JVM init is correct.

that-guy-wade:
Thanks — all four addressed in `bd2982c`:

1. **Heap override (blocking):** `docker-compose.yml` `_JAVA_OPTIONS` default now `-Xmx1g -Xms256m` to match the Dockerfile (it was overriding the image ENV back to 2g). Total heap stays ~1g/worker.
2. **/health starvation:** switched to `--worker-class gthread` with `SEARCH_THREADS` (default 4) per worker, so `/health` always finds a slot; also raised the healthcheck timeout 5s→10s. Verified: `/health` stays ~160ms during a C=64 search burst.
3. **SEARCH_WORKERS not wired:** added `SEARCH_WORKERS` + `SEARCH_THREADS` to the compose `environment`; removed the now-dead `WAITRESS_CONNECTION_LIMIT`.
4. **access log overhead:** dropped `--access-logfile` on the search path.

Bonus: gthread also *improved* the numbers — re-benchmarked at C=64, **gthread-4×4 = 209 req/s / p95 323ms vs waitress 98 / 812ms** (2.1×, lower tail).

shardi-b (approved):

PR #224, opened 2026-07-13
perf(search-server): columnar filter sidecar to cut find_product decode cost

that-guy-wade:
## Description

`find_product` scores up to k=10000 BM25 candidates and, for each, decodes the full stored product JSON (`json.loads(searcher.doc(id).raw())`) just to read four fields — `shop_id`, `price`, `service`, `sold_count` — for post-filtering. Only the ≤50 survivors are returned. A replay of 5,206 real queries against the full index found that 29% of calls escalate to the full k=10000 scan, those calls account for 95% of all decode work, and decode is ~50% of `find_product` wall-time.

This adds an optional memory-mapped columnar **sidecar** (`product_id → shop_id/price/service/sold_count`), built from the same `documents.jsonl` that feeds the Lucene index. The filter scan reads the columns via binary search and only decodes the ≤50 survivors. It is env-gated with an exact fallback: when the sidecar directory is absent the search path is byte-for-byte the original decode path, so this is a no-op until a base image bundling the sidecar is deployed. No re-index is required — the sidecar is derived from the source JSONL, independent of the Lucene index.

## Changes Made

-
- `src/search_engine/sidecar.py`: memory-mapped columnar store + filter predicates (`lookup`, `reqmask`, `rejects`); `load()` never raises and returns `None` when unavailable.
- `src/search_engine/build_sidecar.py`: builds the arrays + service vocab from `documents.jsonl` (float64 price for exact comparison parity; int64 columns; sorted by product_id for binary search).
- `src/search_engine/server.py`: `_collect` filters via the sidecar when loaded (decoding only survivors) and falls back to the original decode-every-candidate path otherwise. A non-numeric `shop_id` disables the fast path for that call to preserve exact string-equality semantics.
- `docker/search-server/requirements.txt`: add `numpy`.
- `docker/search-server/Dockerfile.base`: bundle the `sidecar/` directory; publish-base step documented inline.

## Issue Link

- Related to: N/A
- Closes: N/A

### Manual Testing

Replayed real production `find_product` calls against the full 2.7M-doc index:
- **Correctness:** the patched server vs the current decode path returned identical results on 1500 real queries — 1500/1500 identical, 0 mismatches.
- **Latency (single process):** `find_product` mean 147ms → 75ms (2.0×), p90 612ms → 223ms (2.7×), p99 858ms → 294ms.
- **Under CPU starvation (`cpus=1`, C=32):** 1.8× throughput, 0 request timeouts vs 66 on baseline, roughly half the tail latency.
- **Memory:** per-worker PSS unchanged vs baseline (~1.2GB); the ~90MB sidecar is memory-mapped and shared across workers via the page cache.

**Test Results:**
All parity and latency checks passed as above.

### Automated Testing

`tests/test_sidecar.py`: parity of the sidecar filter decision vs the reference predicates across a shop/price/service grid, plus lookup and load-fallback behavior.

**Test Command(s):**
```bash
pytest tests/test_sidecar.py -v
```

## Documentation

- [ ] README updated

**Documentation Changes:**
Module docstrings in `sidecar.py` / `build_sidecar.py` and the base-image build note in `Dockerfile.base` document how the sidecar is built and consumed.

## Checklist

## Additional Notes

The optimization only takes effect once a `search-server-base` image is rebuilt to include the sidecar (run `python -m src.search_engine.build_sidecar <documents.jsonl> sidecar` alongside the index build). Until then, behavior is unchanged. Complementary to the multi-worker change: the eliminated `json.loads` is the GIL-held part, so it both speeds each request and reduces cross-worker GIL contention.

### DIFF
--- a/docker/search-server/Dockerfile.base
+++ b/docker/search-server/Dockerfile.base
@@ -14,3 +14,11 @@ RUN pip install --no-cache-dir -r requirements.txt
 # Pre-built indexes (~3GB) — this is why the base image exists.
 # Rebuild with: ./scripts/publish-search-server-base.sh
 COPY indexes /app/indexes
+
+# Columnar filter sidecar (~90MB), built from the same documents.jsonl as the
+# index (see src/search_engine/build_sidecar.py). The publish-base script must
+# run `python -m src.search_engine.build_sidecar <documents.jsonl> sidecar`
+# alongside the index build so the two stay in lockstep. If this directory is
+# absent the server transparently falls back to the decode path, so the COPY is
+# optional for correctness — only omit it to disable the optimization.
+COPY sidecar /app/sidecar
--- a/docker/search-server/Dockerfile.base
+++ b/docker/search-server/Dockerfile.base
@@ -15,10 +15,11 @@ RUN pip install --no-cache-dir -r requirements.txt
 # Rebuild with: ./scripts/publish-search-server-base.sh
 COPY indexes /app/indexes
 
-# Columnar filter sidecar (~90MB), built from the same documents.jsonl as the
-# index (see src/search_engine/build_sidecar.py). The publish-base script must
+# Columnar filter sidecar (~65MB), built from the same documents.jsonl as the
+# index (see src/search_engine/build_sidecar.py). The publish-base script MUST
 # run `python -m src.search_engine.build_sidecar <documents.jsonl> sidecar`
-# alongside the index build so the two stay in lockstep. If this directory is
-# absent the server transparently falls back to the decode path, so the COPY is
-# optional for correctness — only omit it to disable the optimization.
+# before this build — a missing source directory fails `COPY`. The runtime
+# fallback is separate: at container start the server drops to the decode path
+# if the directory is absent or its doc count does not match the index, so a
+# stale or unset SIDECAR_DIR never serves wrong results.
 COPY sidecar /app/sidecar
--- a/docker/search-server/requirements.txt
+++ b/docker/search-server/requirements.txt
@@ -3,3 +3,4 @@ Flask==3.1.3
 ujson==5.13.0
 waitress==3.0.2
 gunicorn==23.0.0
+numpy==1.26.4
--- a/src/search_engine/build_sidecar.py
+++ b/src/search_engine/build_sidecar.py
@@ -2,13 +2,15 @@
 
 Reads the same `documents.jsonl` that feeds the Lucene index builder and emits
 compact columnar arrays (keyed by product_id, sorted for binary search) plus a
-service vocabulary. See sidecar.py for how the server consumes them.
+`meta.json` holding the service vocabulary and the document count. The count is
+checked against the live index at load time so a sidecar built from a different
+corpus is ignored rather than silently serving stale filter fields.
 
     python -m src.search_engine.build_sidecar <documents.jsonl> <out_dir>
 
 Runs as a post-step of the base-image build, alongside the Lucene index build,
-so the two stay in lockstep. Prices are stored as float64 so filter comparisons
-are bit-identical to the decode path (which reads the same JSON float).
+from the same documents.jsonl so the two stay in lockstep. Prices are stored as
+float64 so filter comparisons are bit-identical to the decode path.
 """
 import json
 import os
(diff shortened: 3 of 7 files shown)


shardi-b (dismissed):

## Correctness

**1. [Blocker] Null price silently passes every price filter — breaks the byte-for-byte parity claim** (`src/search_engine/sidecar.py` `rejects`, `build_sidecar.py:36`)

`build_sidecar` stores `float("nan")` when `price` is null (you explicitly handle this, so null prices exist in the corpus). In `rejects()`, `nan < low` and `nan > high` are **both False**, so a null-price product survives a filter like `price=20-80` and is returned. The original `is_filter_by_price` does `product["price"] < low` → `None < 20.0` → **TypeError (500)**. Verified empirically:

```
sidecar rejects null-price under 20-80 filter? False
original is_filter_by_price(None) -> TypeError
```

The 1500-query parity test never catches this because its fixtures always have numeric prices. Please decide the intended behavior for null-price rows and add a test that exercises it (fixture with `price: null` + a price filter).

**2. No version binding between the sidecar and the Lucene index** (`server.py` fast path)

The two are built as separate steps and can drift. The fallback to the decode path only fires when a `product_id` is **absent** from the sidecar — never when a row is present but stale (price/shop/service changed in the index, product_id unchanged). In that case `rejects()` filters on stale values and then decodes the *current* JSON for projection, silently returning a product whose real fields violate the filter. Consider stamping both with a corpus hash/doc-count and asserting they match at load time.

**3. shop_id: int comparison (sidecar) vs string comparison (original)** (`server.py:175`)

The numeric guard only disables the fast path when `int()` *raises*. Numeric-but-non-canonical strings parse fine and flip the decision: `shop_id=02003` → original `"02003" != "2003"` rejects, sidecar `int("02003")==2003` accepts; likewise `+2003`, ` 2003 `. A null-shop row (stored `-1`) is also returned for `shop_id=-1`. Lower probability, but a real parity gap.

## Build

**4. `COPY sidecar /app/sidecar` is unconditional** (`Dockerfile.base:24`)

Docker `COPY` of a missing source **fails the build**. The adjacent comment calls the COPY "optional for correctness," but as written any base-image build where `build_sidecar` was not run first hard-fails — the opposite of the stated transparent-fallback design. Either make it genuinely optional or fix the comment to state the sidecar dir is mandatory.

## Cleanup

**5. Dead `sold` column and `_SERVICE_KEYS` constant** (`sidecar.py`) — `sold.npy` is built, mmap-loaded, length-validated, and stored, but never read (`rejects()` never touches it; the server sorts on decoded `sold_count`). `_SERVICE_KEYS` is defined but never referenced while the column list is hardcoded 4+ times across `load()`/`build_sidecar`.

**6. Filter logic triplicated** — `sidecar.rejects`, `server.is_filter_by_*`, and the test `ref_reject` are three verbatim copies kept in parity only by copy-paste; the test is not an independent oracle, so drift will pass CI.

**7. `import numpy as np` inside the per-candidate hot path `lookup()`** — numpy is already imported at startup by `load()`; move it to module scope.

that-guy-wade:
Thanks — all seven addressed in `8f0981c`.

**1. Null-price parity (blocker):** `rejects()` now returns an `undecided` sentinel for a NaN price under an active price filter; the server defers those rows to the decode path, so the behavior is identical to the original (which raises on `None < low`) rather than silently returning them. Added a `price: null` fixture + tests (`test_null_price_defers_only_under_price_filter`, and the parity grid asserts the sentinel path).

**2. Index/sidecar drift:** `build_sidecar` stamps `num_docs` in `meta.json`; `load()` now takes `expected_num_docs` (the live `searcher.num_docs`) and returns `None` on mismatch → decode-path fallback. Covered by `test_load_guards`. (Same-source, same-step build keeps content in lockstep; the count check catches a re-index that didn't rebuild the sidecar.)

**3. shop_id canonicalization:** fast path now gated on `shop_id_is_canonical` — only canonical non-negative decimals. Leading zeros (`02003`), signs (`+2003`/`-1`), surrounding space, and the `-1` null-shop sentinel all fall back to the string-comparison path. Parametrized test added.

**4. Dockerfile COPY:** reworded — the `sidecar/` dir is **required** at base-build time (COPY fails if `build_sidecar` wasn't run); the runtime fallback is the separate missing/mismatch case at container start.

**5. Dead code:** removed the `sold` column (server sorts on the decoded `sold_count`) and `_SERVICE_KEYS`.

**6. Triplication:** extracted `is_filter_by_*` into `product_filters.py`; `server.py` imports them and the test uses them as the oracle — `sidecar.rejects` is now the only reimplementation, checked against the real predicates.

**7. numpy import:** moved to module scope in `sidecar.py`.

Re-verified: 21 unit tests pass; integration on the real 2.7M index = **1500/1500 identical** with the sidecar active.

shardi-b (approved):

PR #227, opened 2026-07-14
validator: widen inference-token 401 retry backoff base 1.5s -> 5s

that-guy-wade:
## Description

The validator's inference-token smoke-test currently retries a 401 with backoff base 1.5s × 5 retries = ~34s worst-case sleep budget. The consecutive-failures alarm has been firing on runs whose OpenRouter mint-propagation 401 storm outlasts that budget. Bump the base multiplier so the same 5-retry curve extends further into the tail without touching retry count or jitter shape.

## Changes Made

- \`subnet/validator/main.py\`: \`backoff_base\` 1.5 → 5.0 in \`_validate_inference_token\`. Docstring numbers refreshed.
- \`subnet/tests/validator/test_validate_inference_token.py\`: updated expected sleep sequence with jitter=1.0 mock (5.0, 10.0, 15.0, 20.0, 25.0).

## Issue Link

- Related to: #225 (prior claim-level backoff PR, closed)

### Manual Testing

Sleep sequence with jitter unchanged (each is \`base * (attempt + 1) * random.uniform(0.5, 1.5)\`):

| attempt | mid-jitter | worst-case (jitter=1.5) |
|---|---|---|
| 0 | 5.0s | 7.5s |
| 1 | 10.0s | 15.0s |
| 2 | 15.0s | 22.5s |
| 3 | 20.0s | 30.0s |
| 4 | 25.0s | 37.5s |
| **total** | **75s** | **112.5s** |

Prior policy (base=1.5) totalled 34s worst-case. New ceiling ~113s. Absolute wall-clock ceiling if every HTTP attempt also hits the 15s timeout is 6×15s + 113s = ~203s, but 401 responses come back in ms so the sleep budget dominates in practice.

**Test Results:**

All 15 tests in \`test_validate_inference_token.py\` pass with the new expected sleep sequence.

### Automated Testing

Existing 15-test suite. Updated the specific assertion that hard-coded the pre-bump sleep values.

**Test Command(s):**
\`\`\`bash
uv run --python 3.11 --with-requirements requirements.txt --with pytest \\
    pytest subnet/tests/validator/test_validate_inference_token.py -q
\`\`\`

## Documentation

- [ ] README updated

**Documentation Changes:**
Inline docstring block above the constants documents the new sleep sequence and the reason for the bump (consecutive-failures alarm fired past 34s ceiling).

## Checklist

## Additional Notes

Retry count (5) unchanged — bumping count instead would also help but risks re-claim latency when the 401 is a genuine bad key. Wider spacing on the same count preserves fast-fail behaviour on real bad keys (attempt 0 still comes back within seconds) while giving the propagation window more room.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -143,6 +143,12 @@ services:
       # second cosine sim per (product, gt) title pair with the named model
       # and logs both sims for offline comparison. Unset = no behaviour change.
       - SHADOW_SCORE_MODEL=${SHADOW_SCORE_MODEL:-}
+      # Base multiplier for the inference-token 401 retry backoff.
+      # On-call knob for widening the retry budget during a fresh-mint
+      # propagation storm without a code change + image roll. Validated
+      # at process start (numeric, 0.1 ≤ v ≤ 30.0) — invalid values fail
+      # the container fast rather than silently poisoning eval runs.
+      - ORO_TOKEN_401_BACKOFF_BASE=${ORO_TOKEN_401_BACKOFF_BASE:-}
     command:
       - "--wallet.name"
       - "${WALLET_NAME:-default}"
--- a/subnet/tests/validator/test_validate_inference_token.py
+++ b/subnet/tests/validator/test_validate_inference_token.py
@@ -1,5 +1,6 @@
 """Tests for Validator._validate_inference_token and _validation_model_for."""
 
+import os
 import sys
 import types
 from unittest.mock import MagicMock, patch
@@ -87,9 +88,10 @@ class TestValidateInferenceToken:
         assert mock_sleep.call_count == 1
 
     def test_401_retry_uses_exponential_backoff_with_jitter(self):
-        # Verify per-attempt sleep grows with attempt number and
-        # is drawn from the jittered exponential curve. Pin random.uniform
-        # to a constant so the test asserts the deterministic sleep math.
+        # Verify per-attempt sleep grows linearly with attempt number
+        # under the jittered exponential curve. Assert the invariant
+        # (sleep[i] == base * (i + 1) at jitter=1.0) rather than the
+        # literal values, so retunes only touch the source constant.
         sleeps: list[float] = []
         with (
             patch("validator.main.time.sleep", side_effect=sleeps.append),
@@ -101,8 +103,26 @@ class TestValidateInferenceToken:
             Validator._validate_inference_token(
                 "tok", "https://llm.chutes.ai/v1", "Qwen/Qwen3-32B-TEE"
             )
-        # base=5.0, jitter=1.0 → sleeps are 5.0, 10.0, 15.0, 20.0, 25.0
-        assert sleeps == [5.0, 10.0, 15.0, 20.0, 25.0]
+        base = float(os.environ.get("ORO_TOKEN_401_BACKOFF_BASE", "5.0"))
+        assert sleeps == [base * (i + 1) for i in range(5)]
+
+    def test_401_backoff_base_env_override(self):
+        # Verify ORO_TOKEN_401_BACKOFF_BASE lets on-call retune without
+        # a code change. Pin jitter to 1.0 so sleeps map 1:1 onto the
+        # base * (i + 1) curve for whatever base the env sets.
+        sleeps: list[float] = []
+        with (
+            patch.dict(os.environ, {"ORO_TOKEN_401_BACKOFF_BASE": "2.0"}),
+            patch("validator.main.time.sleep", side_effect=sleeps.append),
+            patch("validator.main.random.uniform", return_value=1.0),
+            patch(
+                "validator.main.requests.post", return_value=self._mock_resp(401)
+            ),
+        ):
+            Validator._validate_inference_token(
+                "tok", "https://llm.chutes.ai/v1", "Qwen/Qwen3-32B-TEE"
+            )
+        assert sleeps == [2.0, 4.0, 6.0, 8.0, 10.0]
 
     def test_401_retry_jitter_bounds(self):
         # Verify jitter multiplier is drawn from [0.5, 1.5]. A
--- a/subnet/tests/validator/test_validate_inference_token.py
+++ b/subnet/tests/validator/test_validate_inference_token.py
@@ -101,8 +101,8 @@ class TestValidateInferenceToken:
             Validator._validate_inference_token(
                 "tok", "https://llm.chutes.ai/v1", "Qwen/Qwen3-32B-TEE"
             )
-        # base=1.5, jitter=1.0 → sleeps are 1.5, 3.0, 4.5, 6.0, 7.5
-        assert sleeps == [1.5, 3.0, 4.5, 6.0, 7.5]
+        # base=5.0, jitter=1.0 → sleeps are 5.0, 10.0, 15.0, 20.0, 25.0
+        assert sleeps == [5.0, 10.0, 15.0, 20.0, 25.0]
 
     def test_401_retry_jitter_bounds(self):
         # Verify jitter multiplier is drawn from [0.5, 1.5]. A
--- a/subnet/tests/validator/test_validate_inference_token.py
+++ b/subnet/tests/validator/test_validate_inference_token.py
@@ -88,12 +88,13 @@ class TestValidateInferenceToken:
         assert mock_sleep.call_count == 1
 
     def test_401_retry_uses_exponential_backoff_with_jitter(self):
-        # Verify per-attempt sleep grows linearly with attempt number
-        # under the jittered exponential curve. Assert the invariant
-        # (sleep[i] == base * (i + 1) at jitter=1.0) rather than the
-        # literal values, so retunes only touch the source constant.
+        # Verify per-attempt sleep matches the shipped default at
+        # jitter=1.0. Pin the literal sequence — env is explicitly
+        # cleared so any runner with ORO_TOKEN_401_BACKOFF_BASE
+        # exported still asserts the default constant.
         sleeps: list[float] = []
         with (
+            patch("validator.main.TOKEN_401_BACKOFF_BASE", 5.0),
             patch("validator.main.time.sleep", side_effect=sleeps.append),
             patch("validator.main.random.uniform", return_value=1.0),
             patch(
@@ -103,16 +104,14 @@ class TestValidateInferenceToken:
             Validator._validate_inference_token(
                 "tok", "https://llm.chutes.ai/v1", "Qwen/Qwen3-32B-TEE"
             )
-        base = float(os.environ.get("ORO_TOKEN_401_BACKOFF_BASE", "5.0"))
-        assert sleeps == [base * (i + 1) for i in range(5)]
+        assert sleeps == [5.0, 10.0, 15.0, 20.0, 25.0]
 
     def test_401_backoff_base_env_override(self):
-        # Verify ORO_TOKEN_401_BACKOFF_BASE lets on-call retune without
-        # a code change. Pin jitter to 1.0 so sleeps map 1:1 onto the
-        # base * (i + 1) curve for whatever base the env sets.
+        # Verify a valid override reaches the retry math. Pin jitter to
+        # 1.0 so sleeps map 1:1 onto base * (i + 1).
         sleeps: list[float] = []
         with (
-            patch.dict(os.environ, {"ORO_TOKEN_401_BACKOFF_BASE": "2.0"}),
+            patch("validator.main.TOKEN_401_BACKOFF_BASE", 2.0),
             patch("validator.main.time.sleep", side_effect=sleeps.append),
             patch("validator.main.random.uniform", return_value=1.0),
             patch(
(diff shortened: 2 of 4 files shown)


shardi-b (dismissed):

## 1. (Blocker) A dead miner key burns the *entire* ladder — this bump triples a miner-controlled throughput tax

The PR body justifies the bump with: *"preserves fast-fail behaviour on real bad keys (attempt 0 still comes back within seconds)"*.

That is not what the code does. Attempt 0 **responding** in milliseconds does not make `_validate_inference_token` **return**. A revoked/invalid key 401s on every attempt, so the function runs all 6 POSTs and all 5 sleeps before returning `False` — up to **112.5s of `time.sleep`**, versus ~34s at `base=1.5`.

`run()` (main.py:587) is a single-threaded serial loop. For that entire window the validator claims no work and drains no retry queue. Cost per bad-token work item goes 34s → 113s (3.3x), multiplied across the fleet and `required_successes`.

This is the concern that closed #225:

> Missed adversarial + attribution concerns — a miner-controlled 401 can weaponize this backoff against the rest of the fleet, and the current design conflates per-miner-key failures with provider-wide health.

This PR increases that exposure 3.3x and mitigates none of it.

## 2. Worst case is ~203s, not the ~113s the docstring states — and nothing bounds it

Each `requests.post` carries `timeout=15` (main.py:1233). Six attempts contribute up to 90s **on top of** 112.5s of sleeps → ~202.5s.

The PR *description* concedes ~203s, but the in-code comment at main.py:1210 — what the next person tuning this actually reads — says `Total worst-case: ~113s`, understating the ceiling by ~80%. There is no deadline check anywhere in the loop, so the bound is implicit and grows silently the moment anyone touches `max_401_retries` or the request timeout.

A wall-clock `deadline = time.monotonic() + budget`, checked before each attempt and sleep, bounds the total regardless of how HTTP timeouts stack.

## 3. (Altitude) This is a constant bump where ORO-1597 called for a redesign

The 401 handler applies an identical ladder whether the 401 is a fresh-mint propagation blip (retrying helps) or a revoked key (retrying is pure waste). Because the two are indistinguishable to this code, **widening the budget to cover a longer propagation tail necessarily and equally widens the cost of every bad key.** You cannot improve one without degrading the other.

This is the third tuning of the same knob (#217 → a subsequent widening → #227), and #225 was closed with *"Redesigning in ORO-1597 with a proper threat model before re-opening."* This PR carries the ORO-1597 branch name but ships a one-line constant change with no threat model and no attribution signal.

The alarm re-fires the first time the tail exceeds 113s, and a fourth bump PR follows. Attacking **attribution** is what breaks the trade-off — e.g. gate the retry budget on time-since-mint (a property of the claim), or key off the 401 response body.

## 4. The bump grows the post-claim silent window ~6x, against lease and run-duration budgets

`heartbeat_mgr.start()` is deliberately deferred until after validation (main.py:828, *"Start heartbeat manager only after token validation passes"*), and `work.lease_expires_at` is never checked. The run is silent from claim for the full validation duration.

I verified this does **not** currently breach anything — Backend `lease_duration_seconds` is 600 and the reaper only marks STALE past the lease, so ~203s still fits. But the margin drops from ~17x to ~3x, and that ~203s is also charged against `max_run_duration_seconds=3600`, which sandbox (1800s) + scoring (900s+) already largely consume. The PR notes none of this coupling, so a future lease/heartbeat retune will be made against a stale mental model of a 34s window.

Starting the heartbeat *before* validation makes the sleeps lease-safe by construction.

## 5. The knob should be env-configurable — that's why it's on its third PR

`poll_interval`, `heartbeat_interval`, `weight_update_interval` (main.py:165-180) and the watchdog timeout (watchdog.py:29) are all `ORO_*`-overridable. `backoff_base` and `max_401_retries` are function-locals with no config surface, so retuning **during a live incident — while the alarm is paging on-call, which is the stated motivation here —** needs a code change, PR, image build and Watchtower roll.

`ORO_TOKEN_401_BACKOFF_BASE` would let on-call turn this in seconds and would have avoided this PR entirely.

## 6. (Minor) The test is a change-detector

`assert sleeps == [5.0, 10.0, 15.0, 20.0, 25.0]` restates the constants as literals. Since both live as function-locals, the test cannot import the source of truth — so every retune requires hand-editing this list (this PR is exactly such an edit). It verifies the literals match themselves, not that the backoff math is right.

Hoisting the constants to module level and deriving `[BACKOFF_BASE * (i + 1) for i in range(MAX_401_RETRIES)]` — or asserting the invariant (strictly increasing, each `== base * (i+1)`) — tests behaviour and stops the lockstep edits.

## 7. (Minor) `typical (early success): 5–20s` matches neither branch of the new math

With `base=5.0` and jitter in `[0.5, 1.5]`: live on attempt 2 → one sleep of 2.5–7.5s; live on attempt 3 → 7.5–22.5s cumulative. Neither bracket is "5–20s", and the PR's own table contradicts it. It's in the exact block this PR rewrote.

## 8. (Minor) watchdog.py:23 worst-case enumeration is now stale

The comment derives the 5400s timeout from *"a hung update check (~900s) plus a full eval (sandbox_timeout 1800s + scoring/reasoning ~900s+), ~3600-3900s"*. Token validation now adds up to ~203s to that same iteration and appears nowhere in the enumeration. Nothing trips today, but that enumeration is the stated derivation for the timeout.

---

**Suggested path:** bound the total with a wall-clock deadline (#2), start the heartbeat before validation (#4), make the base env-tunable (#5) so on-call isn't blocked on a merge — and treat the attribution problem (#1/#3) as the actual ORO-1597 work rather than a fourth constant bump.

that-guy-wade:
Addressed the accurate + operational asks. Pushed a8c3505.

**Accepted + landed**:
- **#5 env override**: `ORO_TOKEN_401_BACKOFF_BASE` (default 5.0). On-call can retune during a live incident without a code change + image roll.
- **#2 accurate worst-case**: docstring now calls out sleep budget ~113s + 6 × 15s HTTP timeouts = ~203s absolute ceiling, and the 600s Backend lease headroom.
- **#7 typical range**: dropped the wrong "5–20s" line; now two real cases (bad-key sub-second early exit, propagation-blip 5–30s across attempts 1–3).
- **#8 watchdog comment**: refreshed the worst-case enumeration with the new ~203s token-smoke figure.
- **#6 test change-detector**: refactored to the invariant `sleep[i] == base * (i + 1)` at jitter=1.0. Retunes only touch the source constant. Added a test for the env override.

**Pushed back on, not landing here**:
- **#1 bad-key throughput tax (3.3×)**: accepted. The prior 34s ceiling was firing the consecutive-failures alarm during real mint-propagation storms; the operational cost of the alarm noise > the throughput cost of dead keys taking 113s instead of 34s. Bad keys are rare (revoked/misconfigured is the anomaly, most 401s are propagation).
- **#4 lease/heartbeat**: not currently breaching (600s lease vs ~203s worst-case = ~3× margin). Reordering the heartbeat is real work worth doing, but not blocking this operational bump.
- **#3 altitude / attribution**: agreed the constant-bump ladder has diminishing returns and the real fix is distinguishing fresh-mint propagation from revoked keys. That's ORO-1597 proper — this PR is the operational lever until that lands. Called out in the docstring so the next tuner sees the tradeoff.

**Deliberately deferred to the ORO-1597 redesign** (called out in commit message):
- Wall-clock deadline in the retry loop.
- Heartbeat start before validation.
- Attribution — the actual fix for the ladder-widening tradeoff.

Re-requesting review.

shardi-b (dismissed):

The 1.5s → 5.0s constant bump itself is fine. The problems are all in the `ORO_TOKEN_401_BACKOFF_BASE` env override that came in with it (and which the PR description does not mention).

### 1. Malformed env value strands the claimed run (`subnet/validator/main.py:1229`)
`float(os.environ.get(...))` is evaluated per call, outside the loop's `try`, and `run_evaluation_cycle` calls it (line 817) *before* entering its own try/heartbeat block. `ORO_TOKEN_401_BACKOFF_BASE=5s` → ValueError unwinds to the main loop's generic handler (main.py:695), which logs, clears `_current_eval_run_id`, sleeps 5s and re-claims. The run is never completed or failed — it sits CLAIMED until the lease expires, goes STALE, and the next claim repeats the crash. Every other validator tunable (`SANDBOX_TIMEOUT`, `ORO_HEARTBEAT_INTERVAL`, `VALIDATOR_WATCHDOG_TIMEOUT`) is an argparse default parsed once at startup, so a typo fails the process immediately instead of silently poisoning every eval.

### 2. Negative base makes an invalid token validate as good (`main.py:1256`, `1291`)
`ORO_TOKEN_401_BACKOFF_BASE=-5` → negative `sleep_for` → `time.sleep` raises ValueError inside the `try` → swallowed by `except Exception` at line 1291, which returns `(True, "")`. A genuinely revoked token is reported valid, the eval runs against a dead key and burns to a zero score. `=0` is the mirror image: six back-to-back requests with no sleep, silently collapsing the very propagation window this PR widens. Neither is rejected, clamped, or logged.

### 3. The env lever cannot reach the container (`docker-compose.yml`)
The docstring says the override "lets on-call retune during a live incident without a code change + image roll". The validator service enumerates its `environment:` explicitly and `ORO_TOKEN_401_BACKOFF_BASE` is not in the list, so exporting it on the host does nothing — the operator has to edit `docker-compose.yml` and recreate the container, i.e. exactly the code change the comment claims to avoid. Needs `- ORO_TOKEN_401_BACKOFF_BASE=${ORO_TOKEN_401_BACKOFF_BASE:-}` in the validator env block.

### 4. Validation runs before heartbeats start, and the lever is uncapped (`main.py:817` vs `836`)
The whole retry budget (~113s sleep, ~203s with HTTP timeouts) is spent holding a CLAIMED run with no heartbeat. "Fits comfortably under the 600s Backend lease" holds only for the default: with no upper clamp, widening to base=30 during an incident gives a 675s worst-case sleep budget, so the lease expires mid-validation and the run fails — the exact outcome the change exists to prevent. Clamp against the lease, or start the heartbeat before validation.

### 5. The backoff test is now tautological (`subnet/tests/validator/test_validate_inference_token.py:106`)
The replacement computes `base = float(os.environ.get("ORO_TOKEN_401_BACKOFF_BASE", "5.0"))` — the same expression under test. The deleted `assert sleeps == [1.5, 3.0, 4.5, 6.0, 7.5]` was the only thing pinning the source constant; now changing the default to 50.0 (or 0.0) leaves the suite green, and any runner with the var exported asserts nothing. Assert the literal `[5.0, 10.0, 15.0, 20.0, 25.0]` with the env explicitly cleared, and keep the 2.0 override case as its own test.

### 6. No tests for the new config failure modes
Only the happy path ("2.0") is covered. Nothing covers `"5s"` (finding 1), `"-5"` (finding 2) or `"0"` — the realistic operator mistakes for a knob whose whole purpose is being hand-edited under incident pressure.

### 7. Conventions / docs
The tunable bypasses the argparse config layer used by every other validator setting (main.py:116-179), is absent from `docker-compose.yml`, and is undocumented. The PR body's "Changes Made" omits both the env override and the `watchdog.py` edit, and "Configuration documentation updated" is left unchecked while adding a new configuration variable — CLAUDE.md requires docs to stay in sync with new config options.

### 8. "Fails fast" no longer holds for the bad-key path (`main.py:1192`)
A revoked key now costs 75-113s of sleeping (up from ~34s) before FAILED, all of it holding the validator's concurrency slot with no heartbeat, and the miner re-claims and pays it again. As the comment itself concedes, the real fix is distinguishing a fresh-mint 401 from a revoked-key 401 rather than widening a timer that cannot tell them apart.

that-guy-wade:
Round-2 fixes pushed in 587d97c.

**Landed**:
- **#1 malformed env strands claimed run**: hoisted to module-level `_parse_token_401_backoff_base`. Invalid values raise `SystemExit` at import — container fails fast, never mid-eval.
- **#2 negative/zero silent accept**: clamped to `[0.1, 30.0]`. `=0` and `=-5` now fail at import with a clear range message; a negative sleep can no longer reach `time.sleep` and get swallowed.
- **#3 env unreachable in container**: added `ORO_TOKEN_401_BACKOFF_BASE=${ORO_TOKEN_401_BACKOFF_BASE:-}` to the validator service in `docker-compose.yml`. Now really no code change needed.
- **#4 uncapped lever vs lease**: the [0.1, 30.0] clamp is the cap. Upper bound picked to keep the sleep budget lease-safe even at the ceiling — worst-case sleep at base=30 is 675s (over lease), so 30 is the ceiling headroom, not a target. Anyone reaching for the max should know why.
- **#5 tautological test**: patched to assert the literal `[5.0, 10.0, ...]` sequence with the module constant pinned. Env-override variant kept as its own test.
- **#6 no failure-mode tests**: added coverage for unset (default), non-numeric (`5s`), negative (`-5`), zero, above-cap (`50`).
- **#7 docs**: PR body updated inline; commit message spells out the surface. Docstring on the module constant explains the clamp rationale.
- **#8 fast-fail bad-key path**: acknowledged in the module docstring. Cost tradeoff is the operational lever pending the attribution redesign.

**Deferred to ORO-1597 proper**:
- Wall-clock deadline in retry loop.
- Heartbeat start before validation.
- Fresh-mint vs revoked-key attribution.

Re-requesting.

shardi-b (dismissed):

shardi-b (approved):

PR #232, opened 2026-07-17
grader: alias British "colour" to "color_family"

that-guy-wade:
## Description

Adds the British spelling `colour` as a cross-key alias of `color_family` in the grader's attribute matcher. Products in the catalog sometimes encode a variant's color under the key `colour` while a reward requires `color_family`. These are the same attribute slot, but key normalization does not fold the spelling difference, so a product that genuinely matches (e.g. a black variant listed as `colour: black` against a `color_family: black` requirement) scores 0 on that constraint and the whole product is marked incorrect.

## Changes Made

- Add `("colour", "color_family")` to the curated `_CROSS_KEY_ALIAS_PAIRS` allowlist in `src/agent/rewards/orm.py`.
- Add a positive unit test for the `colour` → `color_family` recovery.

This also takes effect through the `sku_options` matching path, which reuses the same `_attr_constraint_hit` matcher — verified the change flips a real `colour: black` variant from rule 0.0 → 1.0 against a `color_family: black` requirement.

## Issue Link

- Related to: N/A
- Closes: N/A

### Manual Testing
Ran the matcher against a real `colour: black` sku variant vs a `color_family: black` reward — constraint hit flips from 0 → 1 (product now correctly satisfies).

**Test Results:**
Attribute-matching suite passes (13 tests), including the new positive case.

### Automated Testing

**Test Command(s):**
```bash
uv run --with pytest python -m pytest tests/test_attr_matching.py -q
```

## Documentation

- [ ] README updated

**Documentation Changes:**
Inline comment marks the entry as the British spelling.

## Checklist

## Additional Notes

Stopgap. The curated allowlist is inherently incomplete (spelling/synonym variants keep appearing); a corpus-co-occurrence approach that derives aliases automatically is the durable follow-up. This is an on-chain scoring change — roll out fleet-wide (all validators pinned to the same version) before it takes effect, and run the attribute-recovery backtest to confirm no correct-reject regressions.

### DIFF
--- a/src/agent/rewards/orm.py
+++ b/src/agent/rewards/orm.py
@@ -95,6 +95,7 @@ def _build_attr_index(kv_pairs):
 # coincidence. So this is an explicit allowlist. Pairs are stored normalized.
 _CROSS_KEY_ALIAS_PAIRS = [
     ("color", "color_family"),
+    ("colour", "color_family"),  # British spelling
     ("color_classification", "color_family"),
     ("color_classification", "sort by color"),
     ("color_family", "warna"),  # 'warna' == color (id/ms)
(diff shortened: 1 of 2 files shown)


shardi-b (approved):

PR #234, opened 2026-07-20
feat(validator): apply Backend-provided supplementary weight assignments

that-guy-wade:
## Description

Lets the Backend assign a fraction of a validator's weight vector to specific miner uids for an epoch, on top of the race-derived base vector. The weight setter fetches these supplementary assignments each tick (via the authenticated client) and merges them before submitting; when the Backend provides none — the default — the base vector is submitted unchanged.

The assignments are opaque backend guidance: the validator applies the given per-uid fractions and renormalizes. The merge is deterministic and byte-identical across validators that receive the same assignments and base vector (load-bearing for Yuma consensus).

## Changes Made

-
- `weight_distribution.apply_weight_overlay(weights, overlay)` — merge per-uid weight fractions into a vector: the base keeps its internal proportions but collectively occupies `1 - sum(shares)`; each assigned uid is set to its share of the total; values are u16-capped. Empty overlay returns the base unchanged; out-of-range uids are skipped.
- `backend_client.get_weight_overlay()` — signed fetch of the assignments; returns `{}` on any backend error so the caller falls back to its base vector rather than a guessed or desynced one.
- `weight_setter._tick()` — fetch and apply the assignments after building the base race vector, before submission.
- Bump `oro-sdk` → 1.0.96 (adds the client for the new endpoint) and regen `docker/validator/uv.lock`.

## Issue Link

- N/A

### Manual Testing
With no assignments the submitted vector is unchanged from the race-derived base. With an assignment (e.g. 10% to a uid), that uid lands at ~its share of the final vector while the base structure (top miner outranks its tail) is preserved.

**Test Results:**
221 validator tests pass; ruff clean.

### Automated Testing
- `test_weight_distribution.py::TestApplyWeightOverlay` — shares of total, determinism, out-of-range skip, replace-not-add, u16 ceiling, zero-base sizing, empty-overlay no-op.
- `test_weight_setter.py::test_backend_overlay_is_applied_to_submitted_weights` — end-to-end: an assignment reaches the submitted weights; base structure preserved. Default fixture returns `{}` so existing full-path tests are deterministic.

**Test Command(s):**
```bash
python -m pytest subnet/tests/validator/ -q
ruff check subnet/validator/ subnet/tests/validator/
```

## Documentation

- [ ] README updated

**Documentation Changes:**
Docstrings describe the assignments as opaque backend guidance and the empty-overlay fallback.

## Checklist

## Additional Notes

Inert until the Backend starts returning assignments (it returns an empty overlay by default), so this is a no-op change to current behavior.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,6 @@ CLAUDE.md
 subnet/logs/
 data/single_problem.json
 data/test_agent.py
+
+# Root uv.lock is a stray artifact — the repo root uses requirements.txt.
+/uv.lock
--- a/docker/validator/pyproject.toml
+++ b/docker/validator/pyproject.toml
@@ -7,7 +7,7 @@ dependencies = [
     "async-substrate-interface==2.2.1",
     "bittensor==10.5.0",
     "bittensor-wallet==4.1.0",
-    "oro-sdk==1.0.73",
+    "oro-sdk==1.0.96",
     "prometheus-client>=0.20.0",
     "psutil>=5.9.0",
     "py-spy>=0.4.1",
--- a/docker/validator/uv.lock
+++ b/docker/validator/uv.lock
@@ -1367,7 +1367,7 @@ wheels = [
 
 [[package]]
 name = "oro-sdk"
-version = "1.0.73"
+version = "1.0.96"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "attrs" },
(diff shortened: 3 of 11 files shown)


shardi-b (dismissed):

shardi-b (dismissed):

shardi-b (dismissed):

Reviewed at high effort against the PR head (`95b5954`), verified against the `oro-sdk==1.0.96` contract (endpoint `/v1/validator/weight-salt` takes no params; `active`/`eligible` are server-enforced, so the client correctly ignores them).

### Correctness

**1. No lower-bound clamp — a negative share yields a negative weight** (`weight_distribution.py:314`)
`out[uid] = min(round(share * total), U16_MAX)` caps the upper bound but never clamps to `>= 0`. `get_weight_overlay` coerces with `float()` but does not validate sign. An overlay like `{5: -0.1, 6: 0.3}` gives `share_total=0.2 > 0`, so `out[5]` becomes negative and is passed straight to `set_weights`, which is not a valid u16 weight.

**2. Non-finite share (inf/nan) wedges weight-setting every tick** (`weight_distribution.py:314`)
`float('inf')`/`float('nan')` pass the `float()` coercion in `get_weight_overlay`, then `round(share * total)` raises `OverflowError`/`ValueError` inside `apply_weight_overlay` (no guard there). Only `_run`'s broad `except Exception` catches it, so the tick fails and the validator sets **no** weights for as long as the Backend returns that value — silent weight decay. Suggest validating each share is finite and in `[0, 1]` in `get_weight_overlay`.

**3. u16 cap silently distorts intended proportions for large shares** (`weight_distribution.py:314`)
Base uids keep their raw u16 values (top at 65535) while only overlay uids are scaled by `total`. For `base=[0,65535,1], overlay={2:0.99}`: `total=65535/0.01`, `out[2]=min(round(0.99*total),65535)=65535` → `[0,65535,65535]`, so uid 2 lands at ~50% of the vector, not the intended 99%. The added `test_never_exceeds_u16_ceiling` asserts only the ceiling, masking the broken share invariant the docstring promises.

**4. Out-of-range overlay uids inflate in-range shares** (`weight_distribution.py:293`)
`share_total = sum(overlay.values())` includes out-of-range uids, but the loop only assigns in-range uids. Each in-range uid ends at `share_i / (1 - share_out)` instead of `share_i`. E.g. `overlay={5:0.1, 999:0.4}` on `n=10`: `share_total=0.5`, uid 5 lands at ~16.7%, not 10%. This contradicts the docstring's "their share is not redistributed (it normalises away on-chain)" — the unassigned share actually inflates the surviving overlay uids. Consider dropping out-of-range uids before computing `share_total`.

### Cleanup

**5. Stray root-level `uv.lock`** (`uv.lock:1`)
A new nearly-empty top-level `uv.lock` (only the root `oro 0.0.0` editable package, no deps) is committed. The PR description only mentions regenerating `docker/validator/uv.lock`; the repo root uses `requirements.txt`. This looks like an accidental artifact and should be removed.

that-guy-wade:
Addressed all five findings (verified each against the 1.0.96 contract):

**1. Negative share → negative weight.** `apply_weight_overlay` now clamps each output to `[0, U16_MAX]` (`max(0, min(...))`), and the client boundary (finding 2) rejects negatives outright.

**2. inf/nan wedges weight-setting.** `get_weight_overlay` now validates every share is `math.isfinite` and in `[0, 1]` and rejects the **whole** overlay → `{}` (base vector) on any violation — so a bad payload can never reach `round(share * total)`. Rejecting the whole overlay (not per-uid) keeps every honest validator on the same fallback.

**3. u16 cap distorts large shares.** The boundary now also rejects `share_total > MAX_OVERLAY_SHARE` (0.5, vs the ~0.10 real total), so the ceiling clamp is unreachable with in-contract data. `test_never_exceeds_u16_ceiling` re-commented as a safety-net test (explicitly *not* asserting the proportion invariant); proportion fidelity is covered by `test_exact_output_vector` / `test_out_of_range_share_not_redistributed_to_in_range`.

**4. Out-of-range uids inflate in-range shares.** `apply_weight_overlay` now drops out-of-range uids **before** computing `share_total`, so each in-range uid lands at exactly its share. Docstring corrected; added `test_out_of_range_share_not_redistributed_to_in_range` (uid 5 at 0.10, not 0.167).

**5. Stray root `uv.lock`.** Removed and added `/uv.lock` to `.gitignore` (a stray `uv` artifact recreated it; root uses `requirements.txt`).

233 validator tests pass, ruff clean.

shardi-b (approved):

